PR #6, opened 2022-12-19
Initial spanish normalization

RaulKite:
Normalization for Spanish Language.
Haven't find a complete spanish.json spelling mappings so only have some examples


RaulKite:
don't merge. working in an improved version

Issue #7, opened 2022-12-20
Does it support Chinese?


tzhgkm:
Hi,

I couldn't find align model used for Chinese in "torchaudio.pipelines". Does it support Chinese?

Thanks

m-bain:
Need a wav2vec2.0 model finetuned on chinese.
Seems there are some on huggingface https://huggingface.co/jonatasgrosman/wav2vec2-large-xlsr-53-chinese-zh-cn

Currently whisperX doesn't support huggingface pipelines though, when I have the bandwidth I can try add this. Otherwise shouldn't be tricky to add this yourself.

m-bain:
Please see recent commit 
https://github.com/m-bain/whisperX/commit/e909f2f766b23b2000f2d95df41f9b844ac53e49
I tried adding this chinese model but it seems not able to align properly. It's hard for me to see what's happening in the dictionary / alignment because im not familiar with the language. @tzhgkm if you can help here that would be great

tzhgkm:
Thank you!
I changed "zh" to "Chinese" in LANGUAGES_WITHOUT_SPACES, and then added Chinese huggingface model.
It seemed that it worked fine for me.

https://user-images.githubusercontent.com/12479016/208854490-a2d6961a-eebc-4f04-8312-9471018ef0c6.mov

mack007liu:
> huggingface模型

huggingface模型 的地址吗？

tzhgkm:
> > huggingface模型
> 
> huggingface模型 的地址吗？

DEFAULT_ALIGN_MODELS_HF = {
    "ja": "jonatasgrosman/wav2vec2-large-xlsr-53-japanese",
    "Chinese": "jonatasgrosman/wav2vec2-large-xlsr-53-chinese-zh-cn",
}

m-bain:
added here (keeping "zh" code) [c6fa7df](https://github.com/m-bain/whisperX/commit/c6fa7df3cc7caae68722363ef0736043f3d75450)

PR #8, opened 2022-12-20
Add Japanese support

yasutak:
# related
#7

# Description

I have implemented support for Japanese by utilizing [jonatasgrosman/wav2vec2-large-xlsr-53-japanese](https://huggingface.co/jonatasgrosman/wav2vec2-large-xlsr-53-japanese) from Hugging Face.

In order to handle the fact that Japanese words are not typically separated by spaces, I break words into individual characters.

An example is as follows. The accuracy is descent.

https://user-images.githubusercontent.com/19920981/208736535-9a32fa18-e0ca-4888-a1ca-9fad506ddbe4.mov

## Type of change

- [x] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

I have tested this implementation by verifying that existing commands are functioning correctly.


m-bain:
Amazing thank you! I made some changes, such as adding default models and type check for huggingface vs torchaudio.

Also, I tried chinese model `jonatasgrosman/wav2vec2-large-xlsr-53-chinese-zh-cn` but it seems not to work. I don't speak so hard to me to test / understand the dictionary

yasutak:
Thank you for merging!

I also attempted to use Chinese and Korean models, but was unsuccessful. I suspect that there may be some bugs in the file `alignment.py`. 

I would be happy to work on adding support for these languages with the assistance of my labmates from Korea and China.

PR #13, opened 2022-12-24
Add Ukrainian wav2vec2 model

egorsmkv:


egorsmkv:
I tested with my model, look at the screenshot below

telegram-cloud-photo-size-2-5276316539898937892-y

Issue #23, opened 2023-01-06
Phoneme-level timestamps


MichaelDoron:
Great work!
Is it possible to expand this to phoneme-level timestamps, instead of word-level timestamps?

For example, instead of 
"[00:13.50->00:13.60] smiles"
have
"[00:13.50->00:13.53] s
[00:13.53->00:13.57] mi
[00:13.57->00:13.58] l
[00:13.58->00:13.60] es"

m-bain:
added in d395c21b8399cb2f29643a75f91469917cdbb991

PR #24, opened 2023-01-07
Add letter timestamps to the output dictionary

MichaelDoron:
Adding timestamps for individual letters in the output of `align` (similar to issue #23 )

Instead of
`{"segments": transcript, "word_segments": word_segments_list}`

align now returns 
`{"segments": transcript, "word_segments": word_segments_list, "letter_segments": letter_segments_list}`


m-bain:
hey thanks for this, sorry i didn't merge because of all the formatting changes on the whole file

PR #28, opened 2023-01-11
Update Portuguese model to wav2vec2-large-xlsr-53-portuguese

rhenanbartels:
Thanks for the amazing work on estimating the time codes of the words. It really helps!

We made some tests with this model and we got better results for Portuguese ASR. Additionally, it is similar to other models available in this module.

https://huggingface.co/jonatasgrosman/wav2vec2-large-xlsr-53-portuguese


rhenanbartels:
@FelippeChemello could you please help validating this hypothesis?

FelippeChemello:
Seems more accurate, I've run here also and produced better results as seen on your benchmark.
For me, go ahead with merge.

PR #34, opened 2023-01-17
Add support to transcribe files within directories

Infinitay:
Adding support to transcribe directories instead of passing in each individual file while retaining the structure. I tested it locally and everything was okay for me, but please test it for yourself just to confirm everything works as intended. Of course, if you have a better solution please let me know or make the changes.

Example command: `whisperx "test dir" "solo_test.wav" --output_dir="test dir output"`

image


Infinitay:
I discovered a potential issue. I'll look into it more and make changes if necessary.

PR #40, opened 2023-01-25
Added arguments and instructions to enable the usage VAD and Diarization

MahmoudAshraf97:
Hello,
I've added a necessary argument to include hugging face access token which is required to load the PyAnnote pipelines


m-bain:
thank you! I realised my token was saved already in a config, so this is great

PR #53, opened 2023-02-01
Add more languages to models list

MahmoudAshraf97:
Hi, I initially added and tested the arabic model by @jonatasgrosman and it worked as expected, and since his models were used in other languages before and I tested the arabic model myself, I assumed it's safe to include all the available models.


maisak3:
thanks, Mahmoud! well when I tried whisperX for Arabic audio, the speaker diarization was incorrect! can you show me an example of how you tried it on Arabic audio?

MahmoudAshraf97:
@maisak3 speaker diarization is not directly related to the alignment model, to test it, use the model to transcribe any arabic audio. Btw mistakes are still expected even in top performing models

PR #57, opened 2023-02-01
support batch processing

TengdaHan:
What it is: Parallalize the `transcribe` function. It gives at least 2X speedup on transcribing.
TODO: supporting temperature sweeping under batch processing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,7 @@ This repository refines the timestamps of openAI's Whisper model via forced alig
 
 <h2 align="left", id="highlights">New🚨</h2>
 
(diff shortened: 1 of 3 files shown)


m-bain:
epic thank you tengda

Barabazs:
Hi @TengdaHan, can you clarify how we should specify an optimal batch size please?

It seems that a batch_size of 16 would divide the audio in 16 equally sized chunks. But does that mean that these 16 chunks will be processed in parallel?

TengdaHan:
@Barabazs The audio file is chunked in the other direction: a batch_size of 16 divides the audio into X batches where each has 16 audio segments.  Then each batch is transcribed together: https://github.com/m-bain/whisperX/blob/main/whisperx/transcribe.py#L415

PR #58, opened 2023-02-01
added turkish wav2vec2 model

MahmoudAshraf97:
tested and it works, this is the best available turkish model on HF, I'm also working on a script/notebook for the community to use in order to test different models and report the WER results and whether it works or not in order to speed up the language addition process


m-bain:
amazing the WER test would be a great addition

PR #74, opened 2023-02-10
added if clause for checking 'level-1'

arnavmehta7:
Fixed the bug which causes failure in alignment when the voice input is too small or has only 1 segment by fixing a value if it doesn't exist.


m-bain (approved):

EgemennSahin:
Hi, is this expected to be merged?

Issue #89, opened 2023-02-17
Memory issues


DaniruKun:
The way that the CLI is implemented currently, it is unusable on a medium-size model and e.g. an hour or two long audio file.
The main problem is that after the transcription, the Whisper model is not properly GCed, and the cache isn't cleared, so there is just not enough memory for both the Whisper and the alignment model.
This is a huge issue since many people have consumer RTX 3000-series cards with only 6GB VRAM.

A quick solution is to not load all models at the same time, and properly GC the whisper model before doing the next steps:
```python
del whisper_model
gc.collect()
torch.cuda.empty_cache()
```

m-bain:
i'll make model flushing an option soon, for cheaper gpu reqs. as well as using faster-whisper (but no ETA yet) as a drop-in replacement for whisper

m-bain:
faster-whisper backend, and model flushing now in v3 version
should reduce gpu mem reqs a lot.
See https://github.com/m-bain/whisperX/issues/159#issuecomment-1521720723 for other ways to reduce gpu mem

PR #91, opened 2023-02-18
Fix memory issues

DaniruKun:
Ref: #89 

This PR fixes some issues with the alignment step.
The alignment model would be loaded too early, resulting in too much CUDA memory usage when both it and e.g. Whisper was loaded.
The solution was to load the alignment model lazily, and garbage collect Whisper and the VAD pipeline before doing any alignment work.

Added new flag `--gc_memory` which will try to gc unusued models and clean CUDA cache, off by default to keep backwards compatibility.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,7 +6,7 @@ from setuptools import setup, find_packages
 setup(
     name="whisperx",
     py_modules=["whisperx"],
-    version="1.0",
+    version="1.1",
     description="Time-Accurate Automatic Speech Recognition using Whisper.",
     readme="README.md",
     python_requires=">=3.7",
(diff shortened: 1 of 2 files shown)


m-bain:
Thanks for this, could you make this memory flush optional with a flag, since there is a tradeoff here of having to reload the models for every new audio file?

DaniruKun:
@m-bain sonething like 'gc-memory'?

m-bain:
> @m-bain sonething like 'gc-memory'?

Sure!

DaniruKun:
@m-bain done, added as `--gc_memory`

DaniruKun:
I will close for now in favor of #169

PR #94, opened 2023-02-20
Make WhisperX depends on openai's whisper lib

Pikauba:
Instead of having copies of openai whisper files we could directly use their lib and benefit from the continual improvements it gets. I changed a few things to the imports to make it work properly. 

Eventually, it could be interesting to refactor the [whisperx transcribe function](https://github.com/m-bain/whisperX/blob/f7093e60d367757e1e32a1da069d7c623455b82d/whisperx/transcribe.py#L23) and wrap around the one in Whisper but for now I let it as is because of few minor differences.

At the moment, I left [assets](https://github.com/m-bain/whisperX/tree/main/whisperx/assets) untouched but it should be removed too as from what I get it will become useless too.


requirements.txt
```diff
@@ -7,3 +7,4 @@ more-itertools
 transformers>=4.19.0
 ffmpeg-python==0.2.0
 pyannote.audio
+openai-whisper
```
slavovthinks, line 10: I would recommend pinning version or commit as they tend to introduce braking changes

arnavmehta7:
Sorry but is your code tested? I think this PR introduces breaking changed because decoding of whisperx is quite different than theirs.

m-bain:
this is added in v2 now :)

PR #103, opened 2023-02-27
FIX: fix VAD for no voice activity less than min_duration_off

Pikauba:
It seems that there is some problems linked to the fact that VAD split the speech into segments that should be considered as a single segment. As described here: [#99](https://github.com/m-bain/whisperX/issues/99) 

Also, isn't it the goal of VAD to only split into segments obvious pauses in the audio (things with pause of more than 1-3s for example)? At the moment, if the model detects a pause of less than a 1s (there is actually no minimum) it will be considered as  if those segments aren't part of the same segment which could lead to bad transcription (splits in middle of sentences).

I changed a few things in the VAD section to make it possible to merge segments of less than an certain amount of time. I placed it in the `min_duration_off ` variable which is available through the `--vad_min_duration_off` option. It is now a tunable option by the user. User can decide if a no voice activity of less than `min_duration_off `should be considered as a real "pause" or not.

@m-bain I am still confused by your [Binarize](https://github.com/m-bain/whisperX/blob/847a3cd85b3e6cf9a7670b6abe3009a3a6a3e611/whisperx/vad.py#L6) implementation and would like to know more about the reason why it actually replaces the one previously used by pyannote [Binarize detection scores using hysteresis thresholding](https://github.com/pyannote/pyannote-audio/blob/b56add2e2033060dca28be71563dab5ff0fd3e64/pyannote/audio/utils/signal.py#L207).
I left it as is in the actual fix and would appreciate the logic behind it!

### DIFF
--- a/whisperx/transcribe.py
+++ b/whisperx/transcribe.py
@@ -1,7 +1,8 @@
 import argparse
+from collections import defaultdict
 import os
 import warnings
-from typing import List, Optional, Tuple, Union, Iterator, TYPE_CHECKING
+from typing import Callable, List, Optional, Tuple, Union, Iterator, TYPE_CHECKING
 
 import numpy as np
 import torch
@@ -14,6 +15,9 @@ from .tokenizer import LANGUAGES, TO_LANGUAGE_CODE, get_tokenizer
 from .utils import exact_div, format_timestamp, optional_int, optional_float, str2bool, interpolate_nans, write_txt, write_vtt, write_srt, write_ass, write_tsv
 from .vad import Binarize
 import pandas as pd
+
+from pyannote.audio import Inference
+from pyannote.core import Annotation
 
 if TYPE_CHECKING:
     from .model import Whisper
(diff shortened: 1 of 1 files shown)


m-bain:
Please see the paper for the reason behind custom binarize implementation:
https://arxiv.org/abs/2303.00747
But you are correct maybe an initial merge (or increasing min_duration_off) would be a good initial step, then dividing as needed.

Pikauba:
Ok! It makes sense with the informations contained in the paper. Indeed as I explained, `--vad_min_duration_off`will let the user choose how to deal with short duration silences and will leave the implementation as it was if --vad_min_duration_off=0.

crisprin17:
The other issue with the implemented Binarize function is that the vad onset and offset are always set to the default.
So, you might want to give the option to change these parameters. 

`def merge_chunks(segments, chunk_size, onset: float = 0.5, offset: Optional[float] = None,):
    """
    Merge operation described in paper
    """
    curr_end = 0
    merged_segments = []
    seg_idxs = []
    speaker_idxs = []

    assert chunk_size > 0
    binarize = Binarize(max_duration=chunk_size, onset=onset, offset=offset)
`

Issue #113, opened 2023-03-04
Hugging_face error with --vad_filter enabled on Windows 10 and pyannote.audio version issue


mshakirDr:
Edit: cannot use `--vad_filter` on Windows. The only way it would work is without this argument, otherwise throws the below error. It works on WLS2 though but the VM gets killed for large audio files so not very helpful there too.

I am running the latest version on Windows with the following arguments (it is a makeshift arrangement to use the cli version in python):
```
device = "cpu" 

sys.argv = ['audio', dir+audio_file,  '--model', 'large-v2', '--align_model', 'WAV2VEC2_ASR_LARGE_LV60K_960H', '--diarize', '--hf_token', token, '--language', 'en', '--parallel_bs', '4',
            '--output_dir', dir, '--fp16', 'False',]# '--vad_filter']
cli()
```
The problem is that when I use `--vad_filter` the first time after installation, it throws the following error:
```
Traceback (most recent call last):
  File "D:\Programming Related\PythonCode\transcription_test.py", line 50, in <module>
    cli()
  File "C:\Users\muham\anaconda3\envs\tt\lib\site-packages\whisperx\transcribe.py", line 643, in cli
    vad_pipeline = Inference(
  File "C:\Users\muham\anaconda3\envs\tt\lib\site-packages\pyannote\audio\core\inference.py", line 99, in __init__
    else Model.from_pretrained(
  File "C:\Users\muham\anaconda3\envs\tt\lib\site-packages\pyannote\audio\core\model.py", line 786, in from_pretrained
    path_for_pl = hf_hub_download(
  File "C:\Users\muham\anaconda3\envs\tt\lib\site-packages\huggingface_hub\utils\_validators.py", line 114, in _inner_fn
    validate_repo_id(arg_value)
  File "C:\Users\muham\anaconda3\envs\tt\lib\site-packages\huggingface_hub\utils\_validators.py", line 172, in validate_repo_id
    raise HFValidationError(
huggingface_hub.utils._validators.HFValidationError: Repo id must use alphanumeric chars or '-', '_', '.', '--' and '..' are forbidden, '-' and '.' cannot start or end the name, max length is 96: 'pyannote\segmentation'.
```
The solution I could find is to first time run the code (as shown above) without the `--vad_filter` argument, let the models downloaded and then re-run the code with this argument.
[...]

PR #114, opened 2023-03-04
Fix hugging face error

mshakirDr:
Model should be loaded with an id to avoid this error: huggingface_hub.utils._validators.HFValidationError: Repo id must use alphanumeric chars or '-', '_', '.', '--' and '..' are forbidden, '-' and '.' cannot start or end the name, max length is 96: 'pyannote\segmentation'.
#113


CristianMaza:
Thanks @mshakirDr 
I couldn't make VAD work because I was getting this error. I checked your pull request and I edited the code manually, now it's working. Not sure why this hasn't been accepted!

PR #122, opened 2023-03-06
Added a function for VAD-segments to handle mp3 files, numpy arrays and tensors

koldbrandt:
I found that the function would let you parse numpy arrays and tensors. The VAD pipeline does however not support this, by just parsing, since you need to parse it as a mapping with the sample rate. 

As of now  you can't parse a mp3 file to the vad pipeline. I fixed this by reading the file with whisper audio_load and then converting it to stereo. This is not a pretty solution, but it seems to work quite well.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,7 +3,7 @@ torch
 torchaudio
 tqdm
 more-itertools
-soundfile
 transformers>=4.19.0
 ffmpeg-python==0.2.0
 pyannote.audio
+soundfile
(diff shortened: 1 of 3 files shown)


sorgfresser:
I'm curious: why do you ensure it's stereo? Doesn't the VAD under the hood use mono as well?

koldbrandt:
> I'm curious: why do you ensure it's stereo? Doesn't the VAD under the hood use mono as well?

As far as I remember then I had some problems when it were mono. It is however almost 2 months ago, and I think Max Bain has already solved this issue with other code.

matheusbach:
Nice improvement. Not sure about stereo

PR #126, opened 2023-03-08
Update diarize.py

ConProgramming:
- Now `assign_word_speakers` also returns chunk level segments split by new speaker
- Now `assign_word_speakers` also returns a set of unique speakers

### DIFF
--- a/whisperx/diarize.py
+++ b/whisperx/diarize.py
@@ -2,7 +2,8 @@ import numpy as np
 import pandas as pd
 
 def assign_word_speakers(diarize_df, result_segments, fill_nearest=False):
- 
+    unique_speakers = set()
+
     for seg in result_segments:
         wdf = seg['word-segments']
         if len(wdf['start'].dropna()) == 0:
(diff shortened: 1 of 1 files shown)


ConProgramming:
I was also going to change `word_segments` to have `"speaker": speaker` and `"text": seg["text"][int(wrow["segment-text-start"]):int(wrow["segment-text-end"])]`... 

But according to the code comment the `.srt` format depends on it being `"text": f"[{speaker}]: " + seg["text"][int(wrow["segment-text-start"]):int(wrow["segment-text-end"])]`?

Barabazs:
Thanks for the contribution, @ConProgramming! The codebase has changed significantly since this was opened. Word-level speaker assignment is now handled via an interval tree in the rewritten diarize.py, and assign_word_speakers returns a single TranscriptionResult with speakers assigned in-place rather than multiple return values.

The one idea here that hasn't been implemented, splitting segments at speaker change boundaries,is still interesting, but would need to be built against the current code. If anyone wants to pick that up, a new PR would be the way to go.

PR #146, opened 2023-03-24
Add Silero VAD as VAD model option; Add Faster-Whisper

cnbeining:
Replace #145 to free up `main` branch of my fork so I can do #121 .

Replace #147 .

Issue #121, opened 2023-03-06
Adds support for FasterWhisper


playdasegunda:
A suggestion would be to add support for [Faster Whisper](https://github.com/guillaumekln/faster-whisper/), which is much faster and uses much less VRAM than Whisper. You can use the Whisper Large V2 model and only use [4.6GB VRAM](https://github.com/guillaumekln/faster-whisper/#large-model-on-gpu) instead of the original 10GB. If you can add this, it will be very useful for the community.

Infinitay:
This is great. I'd love to have this paired in addition to our batch processing for even faster speeds. Does it have any impact on transcription accuracy?

arnavmehta7:
Alright, I researched. Seems like the decoding way, tokenizer of whisper, faster-whisper, and whisperx are quite different.

Barabazs:
Closing because it's already implemented.

PR #147, opened 2023-03-24
Add support for Faster-Whisper with CTranslate2

cnbeining:
Implementing #121 .

Adding support for [Faster-Whisper](https://github.com/guillaumekln/faster-whisper) which changed inference engine to `CTranslate2`. 

This change shall 

- Massive accelerate Whisper on GPU
- Make low-VRAM inference possible (for not super ancient GPUs with `float16` support)
- Made pure CPU inference viable speed-wise (with `int8` for Whisper, and CPU inference on VAD pipeline with Silero VAD waiting to be merged in #146 )

Also address:

- https://github.com/guillaumekln/faster-whisper/issues/39
- https://github.com/openai/whisper/discussions/29#discussioncomment-3726710

CC: @lucasbr15 @arnavmehta7 @acul3 @sailordiary

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,3 +7,4 @@ transformers>=4.19.0
 ffmpeg-python==0.2.0
 pyannote.audio
 soundfile
+faster-whisper[conversion]
--- a/whisperx/__init__.py
+++ b/whisperx/__init__.py
@@ -27,6 +27,30 @@ _MODELS = {
     "large": "https://openaipublic.azureedge.net/main/whisper/models/81f7c96c852ee8fc832187b0132e569d6c3065a3252ed18e56effd0b6a73e524/large-v2.pt",
 }
 
+OPENAI_WHISPER_HUGGINGFACE_MODELS = [
+    "openai/whisper-tiny-en",
+    "openai/whisper-tiny",
+    "openai/whisper-base-en",
+    "openai/whisper-base",
+    "openai/whisper-small-en",
+    "openai/whisper-small",
+    "openai/whisper-medium-en",
+    "openai/whisper-medium",
+    "openai/whisper-large-v1",
+    "openai/whisper-large-v2",
+    "openai/whisper-large",
+]
+
+
+def convert_whisper_to_ct2_model(model_name: str, output_dir: str, quantization: str):
+    """Converts a HuggingFace model to a CTranslate2 model.
+    Ref: https://github.com/guillaumekln/faster-whisper/#model-conversion
+    """
+    print('Converting model to CTranslate2 model...')
+    os.system(f'ct2-transformers-converter --model {model_name} --output_dir {output_dir} \
+    --copy_files tokenizer.json --quantization {quantization}')
+    pass
+
 
 def _download(url: str, root: str, in_memory: bool) -> Union[bytes, str]:
     os.makedirs(root, exist_ok=True)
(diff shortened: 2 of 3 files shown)


arnavmehta7:
WOW

arnavmehta7:
There is one bug - https://github.com/m-bain/whisperX/pull/147/files#diff-cc255fa9b9c90e047cc32c944f047e4b6b72ff5d7aa4a54e66d182716e096c1fR682 
On default, the value of model_dir is None. Hence, when it is unspecified then, it checks os.path.exists(None) which results into a error

cnbeining:
Let me work on that.

cnbeining:
As for decoder logic - YES they are different. 

In this case I opted for a very lazy approach of integration - keeping Faster-Whisper's logic in their package and only take out the `model` which has same output format. Otherwise I will need to create a fork with changes from both this and their repo - and won't have energy to maintain the work.

arnavmehta7:
@cnbeining Then that will cause OOM issues no?

arnavmehta7:
@cnbeining As, I had read. Faster-Whisper uses 30 seconds segments and returns the information according to that, however whisperx maintains very short segments which doesn't cause the aligner (ON GPU) to OOM out. Correct me please if I am thinking wrongly.

arnavmehta7:
> Let me work on that.

In this commit, let's say the user already has this folder in his system while using `plain-whisper` models. In that case, the code will exit out, because the directory will not be empty.

cnbeining:
Unlikely.

- Faster-Whisper takes less VRAM than Whisper as you observed previously
- Silero takes a bit of VRAM and can be ran on CPU only - actually the project was designed to run on single-thread CPU
- Alignment takes another 1.5G of VRAM last time I tested

Worst case Whisper model can be unload from VRAM I believe.

cnbeining:
> > Let me work on that.
> 
> In this commit, let's say the user already has this folder in his system while using `plain-whisper` models. In that case, the code will exit out, because the directory will not be empty.

I have not thought of a reliable way to detect whether directory contains Whisper or Faster-Whisper model: leaving that job to user for now.

arnavmehta7:
> Unlikely.
> 
> * Faster-Whisper takes less VRAM than Whisper as you observed previously
> * Silero takes a bit of VRAM and can be ran on CPU only - actually the project was designed to run on single-thread CPU
> * Alignment takes another 1.5G of VRAM last time I tested
> 
> Worst case Whisper model can be unload from VRAM I believe.

Alignment takes just that amount. Cool, great.

arnavmehta7:
@cnbeining 
```
Downloading: "https://download.pytorch.org/torchaudio/models/wav2vec2_fairseq_large_lv60k_asr_ls960.pth" to /root/.cache/torch/hub/checkpoints/wav2vec2_fairseq_large_lv60k_asr_ls960.pth
100% 1.18G/1.18G [00:04<00:00, 268MB/s]
Performing transcription...
Traceback (most recent call last):
  File "/usr/local/bin/whisperx", line 33, in <module>
    sys.exit(load_entry_point('whisperx', 'console_scripts', 'whisperx')())
  File "/content/whisperX-silero/whisperx/transcribe.py", line 709, in cli
    result = transcribe(model, audio_path, temperature=temperature, **args)
  File "/content/whisperX-silero/whisperx/transcribe.py", line 79, in transcribe
    if model.device == torch.device("cpu"):
AttributeError: 'WhisperModel' object has no attribute 'device'
CPU times: user 215 ms, sys: 37.5 ms, total: 253 ms
Wall time: 28.6 s
```
I was trying to run, but I am getting this error.
```
%%time
!whisperx /content/fg.wav --model large-v2 --align_model WAV2VEC2_ASR_LARGE_LV60K_960H --faster_whisper --model_dir /content/models
```

cnbeining:
Gonna wait for https://github.com/guillaumekln/faster-whisper/pull/70 .

arnavmehta7:
@cnbeining Any other way to run this? I think this is related to `device` attribute not found...
Thanks!

akshhack:
@cnbeining good to go? Is this still active?

invisprints:
I can't wait for the feature too, so I forked the repo and solved some bugs. I have yet to test it with regard to performance. But anyway, it works. You can try it for your test. @arnavmehta7 @akshhack 

https://github.com/invisprints/whisperX

PR #154, opened 2023-03-27
Update decoding.py

FernanOrtega:
New fix for endless loop problem. I also created a PR for official Whisper: https://github.com/openai/whisper/pull/1155

It is explained in https://github.com/openai/whisper/pull/914 and https://github.com/openai/whisper/discussions/924


FernanOrtega:
I think my latest changes don't make sense since the file is removed. 

I also made a PR in Whisper repository so it must solve the endless loop error.

PR #157, opened 2023-03-31
Fix Requirements

Ryan5453:
WhisperX is downloading the PyPi package [whisper](https://pypi.org/project/whisper/) instead of [openai-whisper](https://pypi.org/project/openai-whisper/), which results in an error


Ryan5453:
This should probably be updated to pin to a specific version though in the chance OpenAI makes a breaking change to the API

PR #169, opened 2023-04-09
Optimize the inference process and reduce the memory usage

invisprints:
Inspired by https://github.com/m-bain/whisperX/pull/91.

I have divided the inference loop into three separate loops in order to prevent the simultaneous loading of all the models. This approach will prove to be highly beneficial when transcribing multiple audio files.

### DIFF
--- a/whisperx/transcribe.py
+++ b/whisperx/transcribe.py
@@ -1,5 +1,6 @@
 import argparse
 import os
+import gc
 import warnings
 from typing import TYPE_CHECKING, Optional, Tuple, Union
 import numpy as np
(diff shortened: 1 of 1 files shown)


m-bain:
Hi, this is great but for multiple audio files it means that the models must be reloaded every time.
Thus it is lower memory usage at the cost of higher inference time.

 if you could make this optional with a flag that would be great.

invisprints:
Actually, the model won't reload for each file. The model will transcribe all the files first, then it will be deleted, and we will go to the next stage. That's why I use three loops (vad & transcribe, align and diarize loop) to implement it.

m-bain:
ohhh I understand

PR #178, opened 2023-04-12
Update transcribe.py

SlashHarken:


m-bain:
VAD doesnt work unless .wav format...

PR #180, opened 2023-04-14
add speakers to the text area

invisprints:
If diarize is activated, the speaker is added directly to the text so that all subtitles contain the speaker.

I don't know if this is the best way to handle it, but it achieved my goal at least.


m-bain:
this is done now in the utils file (just not add confusion), speaker key also available

PR #208, opened 2023-05-02
Fixed the word_start variable not initialized bug.

arnavmehta7:


arnavmehta7:
When the condition 
```
                if wrow['start'] != wseg.iloc[wdx]['start']:
``` is not run, then word_start is not allocated. Hence a error is occured.

PR #210, opened 2023-05-02
Update pyannote and torch version

sorgfresser:
This will update `torch` and `torchaudio` to versions `2.0.0` and `2.0.1`. Additionally, we fetch `pyannote` directly from it's development branch - this is currently the only way to enable support for `torch` version 2.
There are a few drawbacks with using the development branch:
- While it is the default branch of `pyannote`, it is not as stable as a release
- Any breaking changes that occur (and they can) will break `whisperX`
- The VAD model is not automatically updated and might not work as well with newer versions of `pyannote`
- We will generate some warnings on loading the VAD model as mentioned in [this issue](https://github.com/pyannote/pyannote-audio/issues/1342) at `pyannote's` repo

So instead of using the development branch, I would propose two other options
1. Use pip to install a specific commit. By doing so we would avoid any breaking changes.
2. Only use pyannote `2.1.1` and it's compatible versions of `torch`, being `1.13.1`. I've made the first commit of this PR specifically for this (5becc99e56e184f2103df2aaf71798626ebdaf46).

Both of them have Pros and Cons too. In any case, using the develop branch of `pyannote` is not a sustainable solution. As soon as there is a release using `torch` version 2, we should use it instead of the develop branch.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,5 @@
-torch==1.13.1
-torchaudio==0.13.1
-pyannote.audio==2.1.1
+torch==2.0.0
+torchaudio==2.0.1
 faster-whisper
 transformers
 ffmpeg-python==0.2.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
-torch==1.11.0
-torchaudio==0.11.0
-pyannote.audio
+torch==1.13.1
+torchaudio==0.13.1
+pyannote.audio==2.1.1
 faster-whisper
 transformers
 ffmpeg-python==0.2.0
--- a/setup.py
+++ b/setup.py
@@ -19,7 +19,7 @@ setup(
         for r in pkg_resources.parse_requirements(
             open(os.path.join(os.path.dirname(__file__), "requirements.txt"))
         )
-    ] + ["pyannote.audio @ https://github.com/pyannote/pyannote-audio/archive/refs/heads/develop.zip"],
+    ] + ["pyannote.audio @ git+https://github.com/pyannote/pyannote-audio@11b56a137a578db9335efc00298f6ec1932e6317"],
     entry_points = {
         'console_scripts': ['whisperx=whisperx.transcribe:cli'],
     },
--- a/setup.py
+++ b/setup.py
@@ -19,7 +19,7 @@ setup(
         for r in pkg_resources.parse_requirements(
             open(os.path.join(os.path.dirname(__file__), "requirements.txt"))
         )
-    ],
+    ] + ["pyannote.audio @ https://github.com/pyannote/pyannote-audio/archive/refs/heads/develop.zip"],
     entry_points = {
         'console_scripts': ['whisperx=whisperx.transcribe:cli'],
     },
--- a/whisperx/asr.py
+++ b/whisperx/asr.py
@@ -181,6 +181,9 @@ class FasterWhisperPipeline(Pipeline):
 
     def preprocess(self, audio):
         audio = audio['inputs']
+        if isinstance(audio, np.ndarray):
+            audio = torch.from_numpy(audio)
+
         features = log_mel_spectrogram(audio, padding=N_SAMPLES - audio.shape[0])
         return {'inputs': features}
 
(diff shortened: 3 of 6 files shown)


sorgfresser:
I thought we have to update the VAD-model at first, turns out the sha you're using matches the up to date one in the `pyannote/segmentation` repo at huggingface.

m-bain:
Thank you!
Just testing this locally with GPU. I agree with forcing a specific commit like https://github.com/m-bain/whisperX/commit/5becc99e56e184f2103df2aaf71798626ebdaf46 until a stable version for torch2.0 is out makes sense.

Also might be worth checking if we can squeeze speed gains with @torch.compile 
https://pytorch.org/tutorials/intermediate/torch_compile_tutorial.html

sorgfresser:
`torch.compile()` is very much unhappy about Numpy-Arrays and the `lru.cache` used in `mel_filters`. How import is that one? Here's the corresponding error

> torch._dynamo.exc.Unsupported: call_function UserDefinedObjectVariable(_lru_cache_wrapper) [TorchVariable(cpu), ConstantVariable(int)] {}

Yes, it is a very descriptive one...

sorgfresser:
Another issue: we're using complex data types for the short-time fourier transfrom. Sadly the default backend for `torch.compile()` does not support complex datatypes. We would have to rewrite this to return real numbers instead and then compute the `abs()` of the two real numbers.

sorgfresser:
Added `torch.compile()` to `log_mel_spectrogram` now. Sadly it can not be applied to most of the other operations as they are fully based on numpy and pandas and `torch.compile()` does not work with them (at least not yet). The `fullgraph` keyword is not required, so we could drop it, but it works with it and if I got the Pytorch Tutorial correctly it should speed it up a bit.

youssefabdelm:
I'm not sure if this is related to your changes @sorgfresser , I tried cloning the repo, and merging your changes, then pip installing from that clone. I'm getting this error when transcribing:

```
INFO:pytorch_lightning.utilities.migration.utils:Lightning automatically upgraded your loaded checkpoint from v1.5.4 to v2.0.2. To apply the upgrade to your files permanently, run `python -m pytorch_lightning.utilities.upgrade_checkpoint --file ../root/.cache/torch/whisperx-vad-segmentation.bin`
Model was trained with pyannote.audio 0.0.1, yours is 2.1.1. Bad things might happen unless you revert pyannote.audio to 0.x.
Model was trained with torch 1.10.0+cu102, yours is 2.0.0+cu118. Bad things might happen unless you revert torch to 1.x.
Transcribing...
[2023-05-03 23:15:17,142] torch._dynamo.convert_frame: [WARNING] torch._dynamo hit config.cache_size_limit (64)
   function: 'log_mel_spectrogram' (/usr/local/lib/python3.10/dist-packages/whisperx/audio.py:88)
   reasons:  padding == 84180
to diagnose recompilation issues, see https://pytorch.org/docs/master/dynamo/troubleshooting.html.
---------------------------------------------------------------------------
Unsupported                               Traceback (most recent call last)
[<ipython-input-14-bc55d3217283>](https://localhost:8080/#) in <cell line: 71>()
     69 model = whisperx.load_model(model.strip().lower(), device)
     70 print('Transcribing...')
---> 71 result = model.transcribe(audio_file)
     72 print('Completed transcribing...')
     73 

12 frames
/usr/local/lib/python3.10/dist-packages/torch/_dynamo/exc.py in unimplemented(msg)

Unsupported: cache_size_limit reached
```

m-bain:
@youssefabdelm interesting, do you have enough GPU mem?

So for me locally this PR works, however when I try on google colab the transcription stage hangs forever... (maybe its not using gpu correctly)

sorgfresser:
@youssefabdelm pretty sure it is caused by me. How long was the audio file you wanted to transcribe? The padding seems to be a problem. Could you provide me an example to reproduce?

sorgfresser:
@m-bain interesting! Just ran it in google colab w GPU using `large-v2` and it worked for some 6min audio file. I am concerned whether it's related to the length (as `cache_size` would indicate).

sorgfresser:
Tracked down the root cause of the stalling in colab to `pyannote.audio` not doing the diarization properly. I opened an issue over there (pyannote/pyannote-audio#1360).

m-bain:
I remove the compile statements, since this was breaking some environments I had and wasn't giving much speedup, will return to torch.compile stuff down the road -- although I dont expect huge gains

m-bain:
Im also still finding google colab to run super slow, like 5+ mins for 25min of audio, can someone share colab thats working fast if they manage to

sorgfresser:
I can share you one but I'm not sure whether it's fast anymore since you removed the `device=device` keyword from the diarization pipeline. This is sadly needed, as pyannote/pyannote-audio#1354 is not fixed yet.
Can not test if the difference to the version with `device` keyword is as big as I think it is since I reached my limit for the day. Still, there you go so you can test for urself. [https://colab.research.google.com/drive/1uMWGutVsO6XAN0OvFPOcds0dcepYfaJX?usp=sharing](https://colab.research.google.com/drive/1uMWGutVsO6XAN0OvFPOcds0dcepYfaJX?usp=sharing)

m-bain:
oops, yea my vscode checked out the wrong thing, thanks I added that back

PR #217, opened 2023-05-04
Remove hyphen from dict keys

sorgfresser:
Substitutes every hyphen in a dictionary key with an underscore.
Why is this needed, isn't it just a question of code style? Hyphen is a reserved character and as such indexing with it will not work. This is usually not an issue as one can simply take quotation marks, but this is not always possible. This problem occurred to me when I was creating `TypedDict` classes out of the `word_segments` dictionaries.
To be precise, the following code failed
```
class SingleWordSegment(TypedDict):
    start: float
    end: float
    text: str
    word-segments: str # this is not a str, just as a dummy
    char-segments: str # this is not a str, just as a dummy
```
Neither escaping nor quotation marks are possible in such a case as `TypedDict` requires a Python identifier. The only way to fix this is removing the hyphen, this PR achieves it.

### DIFF
--- a/whisperx/alignment.py
+++ b/whisperx/alignment.py
@@ -140,9 +140,9 @@ def align(
     prev_t2 = 0
 
     char_segments_arr = {
-        "segment-idx": [],
-        "subsegment-idx": [],
-        "word-idx": [],
+        "segment_idx": [],
+        "subsegment_idx": [],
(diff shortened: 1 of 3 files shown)


m-bain:
thanks, yes dict keys no longer have this :)

Issue #220, opened 2023-05-06
Add progress callback


weisrc:
It will be great to know the progress of the transcribe step and align step as a callback that is passed to it if the user wants the progress information.

Barabazs:
Added in #1371

PR #222, opened 2023-05-06
add on_progress callback

weisrc:
This addresses issue #220.


weisrc:
However, this is only done on V3.

PR #303, opened 2023-06-05
Suppress numerals

m-bain:
- Option to suppress numerals at whisper transcription stage
- Ensures speech "2017" is transcribed as "twenty seventeen", or "two zero one seven".
- This helps alignment stage since numeric words cannot be aligned
- Add `--suppress_numerals` flag to CLI, or possible to add to python API (using asr_options dict)


arnavmehta7:
Does this causes any tradeoff with normal transcription?

heilrahc:
how to use this in python?

PR #309, opened 2023-06-07
Add Audacity export

Ca-ressemble-a-du-fake:
Hi,

I wanted to check if the transcript produced by WhisperX had improved for French compared to the first time I tested it. And it has ... tremendously! This works really great. 

So I needed to check it in Audacity, that's why I made this export and I propose it now to the community. 

This exports the transcript to a text file that can be directly imported in Audacity as label file. This is useful to visually check the transcript-audio alignment.

To use it, fire up Audacity, then do "Import Labels" with this file.

WhisperX_Audacity2


m-bain:
Nice work, thank you! I just made some small changes to make the `.aud` format an optional export format, since dont want to save too many files by default

Issue #340, opened 2023-06-23
Running with different parameters gives identical results


isadielen:
Hi,

I am running some experiments with WhisperX in which I change parameters in the function whisperx.load_model() such as `temperature`, `vad_onset`, or `offset`. 

However, this is giving me identical results each time. It looks like it is not using the parameters in `asr_options` and `vad_options` within the load_model() function. The parameters `default_vad_options` and `default_asr_options` get updated in `asr.py`, but it has no influence on the transcriptions.

Anyone an idea how to solve this problem?

Thank you!

anouk1512:
I encountered the same problem! No idea what is happening...

Jeronymous:
Same here.
I wanted to vary `--vad_onset` and `--vad_offset` options, but whatever their value is, the `Binarize` class is always instantiated with default values : onset = 0.5 and offset = None (meaning offset = onset = 0.5).
The __init__ of this class being here https://github.com/m-bain/whisperX/blob/befe2b242eb59dcd7a8a122d127614d5c63d36e9/whisperx/vad.py#L94
(Note: the values of vad onset/offset are actually taken into account to build the VAD pipeline (VoiceActivitySegmentation), but it does not have any effect there)

So even the default value --vad_offset of the CLI `(default: 0.363)` seems to be wrong (it's actually 0.5 that is used)

ricardojeq:
I tried adding the faster_whisper's implementation of temperature under the function generate_with_fallback().

Here is my version :

   def generate_segment_batched(self, features: np.ndarray, tokenizer: faster_whisper.tokenizer.Tokenizer, options: faster_whisper.transcribe.TranscriptionOptions, encoder_output = None):
        batch_size = features.shape[0]
        all_tokens = []
        prompt_reset_since = 0
        if options.initial_prompt is not None:
            initial_prompt = " " + options.initial_prompt.strip()
            initial_prompt_tokens = tokenizer.encode(initial_prompt)
            all_tokens.extend(initial_prompt_tokens)
        previous_tokens = all_tokens[prompt_reset_since:]
        prompt = self.get_prompt(
            tokenizer,
            previous_tokens,
            without_timestamps=options.without_timestamps,
            prefix=options.prefix,
        )

        def decode_batch(tokens: List[List[int]]) -> str:
            res = []
            for tk in tokens:
                res.append([token for token in tk if token < tokenizer.eot])
            # text_tokens = [token for token in tokens if token < self.eot]
[...]

PR #357, opened 2023-07-06
add temperature loop with fallback and pass **kwargs to model.generate

Boulaouaney:
solves #340


ricardojeq:
Thanks for using my code.

Issue #200, opened 2023-04-30
V3 sentence segement issue


seset:
V3 is now incrediably fast, maybe dozens of times faster  
but now the subtitles of each paragraph are too long . examples as below :
1
00:00:00,730 --> 00:00:26,190
Are you nervous? It's a good nervous, a happy nervous. Yeah, Matt's an incredible man, and it's obvious he's very much in love with you. I know. He's a little bit nervous too, you know, but he's holding up great. I'm glad. Yeah. Okay, well, I'm gonna go downstairs and work on those decorations. If you need anything, let me know. Love you, Mom. Love you, sweetheart. Oh, thank you.

2
00:00:34,973 --> 00:01:02,883
Oh, Jesus Christ. I thought you'd never leave. You need to get out of here, Ben. Like, right now. She could have caught us. But she didn't. And we weren't doing anything anyway. And who cares even if we were? I care? Okay, I'm not falling for one of your I wanna be with you routines. You only want me because you can't have me. Yeah, and you're only resisting because you're too preoccupied with being a petty rule follower. I am not.

m-bain:
Add the following

--max_line_width 42 --max_line_count 2

seset:
> Add the following
> 
> --max_line_width 42 --max_line_count 2

thanks ,it can somehow impove...and when i try japanese or chinese file source, there is a lot repetitive words/letters/characters in srt and vtt output, while the json and txt content are correct,  a chinese example mixed with english as below for your easying understanding what went wrong.  Since the JSON file is correct, hopefully this minor error can be fixed soon:)

correct content(json and txt):-----------------------------------------------------
[JSON]
[segments]
[0]
[start : 0.6180930656934307]
[end : 22.254860401459855]
[text : "如何巧妙的去使用多个不同的ControlNet并用到Stable Diffusion的一些最新插件去精准的控制画面 生成一些这样的图片我研究了一整个星期 尝试了MultiControlNet不同的排列组合调整了很多不同的参数 把坑都踩过了一遍给你们整理了12种高阶用法的合计和所需要的所有参数"]
[words]

wrong and repetitive (srt,vtt):------------------------------------------------------
1
00:00:00,618 --> 00:00:22,235
[...]

PR #445, opened 2023-08-29
feat: Add merge chunks chunk_size as arguments.

jim60105:
Suggest from https://github.com/m-bain/whisperX/issues/200#issuecomment-1666507780

This PR add a argument `chunk_size` with `int` input
Default is 30, reduce this if the chunk is too long.

```
whisperx --chunk_size 30 audio.wav
```

The length of sentences in different languages varies. I believe it is reasonable to extract the chunk_size as a parameter. 

In the case of Chinese and Japanese, the suggested value of 5~10 (by @seset) is relatively appropriate. You can clearly see the difference in the following image.

| chunk size 6 | chunk size 30 |
|-|-|
|||

### DIFF
--- a/whisperx/asr.py
+++ b/whisperx/asr.py
@@ -247,7 +247,7 @@ class FasterWhisperPipeline(Pipeline):
         return final_iterator
 
     def transcribe(
-        self, audio: Union[str, np.ndarray], batch_size=None, num_workers=0, language=None, task=None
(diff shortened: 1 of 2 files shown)


arlovibre:
When I try to use this parameter, I am getting unrecognized arguments: --chunk_size error. Can you help me please?

PR #459, opened 2023-09-04
A solution to long subitles and words without timestamps

awerks:
Example usage: 
subtitles_proccessor = SubtitlesProcessor(output["segments"], detected_language, max_line_length = 50, min_char_length_splitter = 35) subtitles_proccessor.save("subtitles.srt", advanced_splitting = True)


m-bain:
You could add example to readme

PR #507, opened 2023-10-05
Fix: Allow vad options to be configurable by passing to FasterWhisperPipeline and merge_chunks.

abettke:


abettke:
Shoutout to my colleague @crisprin17, the actual brains behind these changes.

PR #522, opened 2023-10-12
Update the list of supported phoneme ASR alignment languages.

xennygrimmato:
The list of supported phoneme ASR alignment models can be found here: https://github.com/m-bain/whisperX/blob/main/whisperx/alignment.py#L24-L54


deep-pipeline:
@m-bain This is an uncontroversial README improvement suggestion but the language list update is now a little out of date (from Oct 2023) if one looks at the relevant section of code linked to above..

PR #757 is probably a better approach and if incorporated would render this PR worth closing un-actioned. It might be encouraging of contributions to merge this and then immediately merge #757 (or else just reject this here and leave a thank you note for the original contributor?).

Barabazs:
Thank you for your contribution.
I'll be closing this in favor of a more recent PR.

PR #529, opened 2023-10-15
remove the minimum length for alignment and print the failing segment

MahmoudAshraf97:
In a #510 I fixed the minimum input length problem and so I don't see a reason why we can't align short segments that are often consisting of a single word


m-bain:
Fair, I sometimes found the ultra short segments were hallucination (before the VAD feature was added), but I'll add this, seems less of an issue now with vad and things

Issue #540, opened 2023-10-24
pyannote version 3.0.1 breaks MacOS compatibility


daeh:
The pyannote dependency was bumped to `3.0.1` in d4ac9531d90932d514e5401c892a81ee6674f176 

This forces a `onnxruntime-gpu` dependency rather than `onnxruntime`, which is not compatible with Mac hardware.

See issues on pyannote https://github.com/pyannote/pyannote-audio/issues/1505
and on onnxruntime https://github.com/microsoft/onnxruntime/issues/12252

pip install error:
```shell
# > pip install git+https://github.com/m-bain/whisperx.git
INFO: pip is looking at multiple versions of pyannote-audio to determine which version is compatible with other requirements. This could take a while.
ERROR: Could not find a version that satisfies the requirement onnxruntime-gpu>=1.16.0 (from pyannote-audio) (from versions: none)
ERROR: No matching distribution found for onnxruntime-gpu>=1.16.0
```

justinwlin:
+1 added to a different issue that I had to make a Docker container with the setup.py reverted to 3.0.0 so that I could get it working. 

https://github.com/m-bain/whisperX/issues/499

Barabazs:
Closing since this seems to be resolved by now.

PR #541, opened 2023-10-24
Update setup.py to download pyannote depending on platform

justinwlin:
https://github.com/m-bain/whisperX/issues/540

Breaks Mac platform and arm64 platforms. 

Tested by successfully building on my m1 mac doing:
pip install .

And also using Depot (a way to build Dockerfiles remotely, using arm64):
depot build -t whisperx_image . --platform linux/arm64 --load

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,7 +1,15 @@
 import os
-
+import platform
 import pkg_resources
 from setuptools import setup, find_packages
+
+
+def get_pyannote_audio_version():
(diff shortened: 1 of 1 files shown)


m-bain:
thank you, yes sorry about the mac issues. I had originally been using this for desktop / commerical gpus, since thats quite common in academia. But seems a lot are using it on mac

PR #548, opened 2023-10-29
Add new option to generate subtitles by a specific number of words

amolinasalazar:
Added a new word option called **--max_words_count** that will generate subtitles setting a maximum limit of words per segment.

I've opened the same PR in Whisper: [https://github.com/openai/whisper/pull/1729]() with some slightly changes:

- All options are compatible: **"highlight_words", "max_line_count", "max_line_width" and "max_words_count"**.
- Tested with/without diarization.
- Please pay attention to the changes when I handle "times" in iterate_subtitles() on util.py file. I've wanted to minimize the impact on the current behavior, but I changed the times we yield so now they match the chunk of words set instead of the original segment.

The rest is just a copy/paste of the original PR description:

Added a new word option called **--max_words_count** that will generate subtitles setting a maximum limit of words per segment. This could sound similar to **--max_line_width** option, but the results are more pleasent for readers IMHO. Here a couple of comparisons using .SRT files:

max_word_count
Notice that **--max_words_count** works as an upper bound of words, but still it will respect the segments in the way that end of sentences can have less words if the remaining number of words in a segment is lower than the max_words_count value.
i.e. _Segment = [word1, word2, word3, word4, word5] and max_words_count = 3
=>Result = [word1, word2, word3] and [word4, word5]_
This is not the behaviour we can see using **--max_line_width** that can leave bigger gaps of time when joining end and beginning of segments:
WidthvsWords

Subtitles generated with **--max_words_count** look similar of what we can see in Shorts, Reels and other short duration videos.

This is my first contribution, so feel free of changing/comment/improve anything.

### Additional notes

- Manually tested using Python and cli and checked results in .srt and .vtt files (.txt. and .tsv files won't be affected).

This is my first contribution, so feel free of changing/comment/improve anything.

### DIFF
--- a/whisperx/transcribe.py
+++ b/whisperx/transcribe.py
@@ -69,6 +69,7 @@ def cli():
     parser.add_argument("--max_line_width", type=optional_int, default=None, help="(not possible with --no_align) the maximum number of characters in a line before breaking the line")
     parser.add_argument("--max_line_count", type=optional_int, default=None, help="(requires --no_align) the maximum number of lines in a segment")
     parser.add_argument("--highlight_words", type=str2bool, default=False, help="(requires --word_timestamps True) underline each word as it is spoken in srt and vtt")
+    parser.add_argument("--max_words_count", type=optional_int, default=None, help="the maximum number of words in a segment")
     parser.add_argument("--segment_resolution", type=str, default="sentence", choices=["sentence", "chunk"], help="(not possible with --no_align) the maximum number of characters in a line before breaking the line")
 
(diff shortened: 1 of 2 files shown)


amolinasalazar:
The main PR got merged in Whisper: [https://github.com/openai/whisper/pull/1729](https://github.com/m-bain/whisperX/pull/548) I'll apply the latest changes.

amolinasalazar:
Merged conflicts and bringed the new naming from Whisper:

- RENAME from max_words_count to max_words_per_line: https://github.com/m-bain/whisperX/pull/548/commits/6ba8d6af6c8b0a44d806155081933b8ac6c3862b

Again, feel free to review/test/change anything.

PR #549, opened 2023-10-31
Minor fixes for word options and subtitles

amolinasalazar:
Some minor fixes I noticed when I was making changes for [https://github.com/m-bain/whisperX/pull/548]()

- Removed references to **--word-timestamps** as it's not a valid option in WhisperX. Rephrased to **(not possible with --no_align)** as word options need the word segments that are generated when doing the alignment. **PLEASE CORRECT ME IF I'M WRONG.**
- Removed some duplicated code.


m-bain:
its true, thanks

PR #552, opened 2023-11-02
feat(align-model-hf): add default Romanian(ro) default align model in HuggingFace

fseasy:
## What's

I found a Romanian align model from hugging face: https://huggingface.co/anton-l/wav2vec2-large-xlsr-53-romanian
It has over 1K downloads and I test it in one sample audio, it works.
So I put it here if you need this.


Majdoddin:
RP #791  uses a model for Romanian that has better evaluations.

PR #563, opened 2023-11-07
feat: Load audio from buffer or filename path

mclaudiu23:
Hello,

I think it is interesting to have the option to read from a buffer and not just from the file path.
I have tested the solution and it works correctly.

Thanks


Barabazs:
Please include a usage example.

PR #577, opened 2023-11-15
large-v3 support

gorkemgoknar:


gorkemgoknar:
Should be a draft branch as it may break other whisper models

zamoshchin:
AttributeError: 'WhisperTokenizer' object has no attribute 'decode_batch'

PR #585, opened 2023-11-16
Add alignment model for Malayalam

kurianbenoy:
- Added the phoneme based model `https://huggingface.co/gvs/wav2vec2-large-xlsr-malayalam`, which is most popular wav2vec model available in huggingface


m-bain:
did you test it qualitatively?

kurianbenoy:
Yes @m-bain. It is doing a pretty decent job in transcribing long form malayalam speech. One tweak though I had leverage was to use fine-tuned whisper model weight to get good performance as native performance of Whisper is not so good in Malayalam.

https://github.com/kurianbenoy/malayalam-whisperX/blob/main/malayalam_whisperx.ipynb

PR #586, opened 2023-11-17
Update pyannote to 3.1.0

### LINKED ISSUE #499: pyannote/speaker-diarization-3.0 runs slower than pyannote/speaker-diarization@2.1
https://github.com/m-bain/whisperX/blob/07fafa37b3ef7ce8628b194da302a5a996bb7d37/setup.py#L22
Currently pyannote.audio is pinned to 3.0.0, but it has been [reported](https://github.com/pyannote/pyannote-audio/issues/1481) that it performed slower because the embeddings model ran on CPU. As a result a new[ release 3.0.1](https://github.com/pyannote/pyannote-audio/pull/1478) fixed it by replacing`onnxruntime` with `onnxruntime-gpu`.

It makes sense for whisperX to update pyannote.audio to 3.0.1, however, there is a conflict with faster_whisper on `onnxruntime`, as discussed [here](https://github.com/guillaumekln/faster-whisper/issues/493). Until it is resolved on the faster_whisper side, installing both will end up `onnxruntime` still in CPU mode and thus slower performance.

My current workaround is running the following commands post installation
```sh
pip install pyannote.audio==3.0.1
pip uninstall onnxruntime
pip install --force-reinstall onnxruntime-gpu
```

Alternative, use the old 2.1 model.
```py
model = whisperx.DiarizationPipeline(model_name='pyannote/speaker-diarization@2.1', use_auth_token=YOUR_AUTH_TOKEN, device='cuda')
```

remic33:
Close #499 
Pyannote now fully can work with faster_whisper and on GPU
Do not forget to ask permission for new weights access on hunggingface : https://huggingface.co/pyannote/speaker-diarization-3.1

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,15 +1,8 @@
 import os
 import platform
+
 import pkg_resources
-from setuptools import setup, find_packages
-
-
-def get_pyannote_audio_version():
-    machine = platform.machine()
-    system = platform.system()
-    version = "3.0.0" if machine == "aarch64" or system == "Darwin" else "3.0.1"
-    return version
-
+from setuptools import find_packages, setup
 
 setup(
     name="whisperx",
@@ -28,7 +21,7 @@ setup(
             open(os.path.join(os.path.dirname(__file__), "requirements.txt"))
         )
     ]
-    + [f"pyannote.audio=={get_pyannote_audio_version()}"],
+    + [f"pyannote.audio==3.1.0"],
     entry_points={
         "console_scripts": ["whisperx=whisperx.transcribe:cli"],
     },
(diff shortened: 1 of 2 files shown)


stevenpetryk:
Should the blurb in the README that says the following be removed? I think it'd be considered misleading at this point.

> As of Oct 11, 2023, there is a known issue regarding slow performance with pyannote/Speaker-Diarization-3.0 in whisperX. It is due to dependency conflicts between faster-whisper and pyannote-audio 3.0.0. Please see https://github.com/m-bain/whisperX/issues/499 for more details and potential workarounds.

PR #601, opened 2023-11-26
Add on_progress callbacks to align() and transcribe()

### LINKED ISSUE #220: Add progress callback

stex:
This PR adds an optional `on_progress` argument to both the `align` and `transcribe` methods. 

This allows processing the current progress via a callback, until now, it was only possible to print the progress to `STDOUT`.

Example:

```python
transcribe(
  my_audio, 
  batch_size=8,
  on_progress=lambda state, current=0, total=0: print(f"state: {state}, current: {current}, t: {total}")
)
```

States are defined as

```python
class TranscriptionState(Enum):
  LOADING_AUDIO = "loading_audio"
  GENERATING_VAD_SEGMENTS = "generating_vad_segments"
  TRANSCRIBING = "transcribing"
  FINISHED = "finished"
```

Closes #220


stex:
@m-bain What do you think of the change? It should be backwards-compatible and makes it easier to use `whisperx` as a library

PR #636, opened 2023-12-19
Adding Norwegian Bokmål and Norwegian Nynorsk

peregilk:
Adding Wav2Vec2-models for Norwegian Bokmål and Norwegian Nynorsk. The models are testet together with WhisperX, and works great. For Bokmål I have added the 1B model, even if I see fairly little difference between that and the 300M model. For Norwegian Nynorsk only a 300M model exist. The quality of the Wav2Vec models are also reported here: https://arxiv.org/abs/2307.01672


peregilk:
Just realised that we also have a 1B Nynorsk model. Maybe update the address to this:
NbAiLab/nb-wav2vec2-1b-nynorsk

m-bain (approved):

PR #646, opened 2023-12-26
Update pyannote to v3.1.1 to fix a diarization problem (and diarize.py) 

santialferez:
Hi!

I've update the `setup.py` to include the new version of `pyannote.audio==3.1.1`. That fix a problem related to https://github.com/pyannote/pyannote-audio/issues/1567. I've also update `diarize.py` to include the parameter `num_speakers`. 

Best,


jimmy6DOF:
could also be good time to remove the below quoted note on this from Readme (103-104):

> Note
> As of Oct 11, 2023, there is a known issue regarding slow performance with pyannote/Speaker-Diarization-3.0 in whisperX. It is due to dependency conflicts between faster-whisper and pyannote-audio 3.0.0. Please see https://github.com/m-bain/whisperX/issues/499 for more details and potential workarounds.

PR #657, opened 2024-01-04
Fix VAD Path for Custom VAD loading

Swami-Abhinav:
Fix VAD Path for Custom VAD loading


3manifold (changes requested):
`vad_model` variable  is used to manually   assign a  `vad`   method  that    inherits from  `Vad` class (see `def load_model` in `whisperx/asr.py`). Thus, there is no _fix_ that makes sense in this case.

PR #620, opened 2023-12-09
Add on_progress callback for transcribe and align

stex:
This adds an optional `on_progress` argument to both the `align` and `transcribe` methods.

This allows processing the current progress via a callback, until now, it was only possible to print the progress to `STDOUT`.

Example:

```python
transcribe(
  my_audio,
  batch_size=8,
  on_progress=lambda state, current=0, total=0: print(f"state: {state}, current: {current}, t: {total}")
)
```

States are defined as

```python
class TranscriptionState(Enum):
  LOADING_AUDIO = "loading_audio"
  GENERATING_VAD_SEGMENTS = "generating_vad_segments"
  TRANSCRIBING = "transcribing"
  FINISHED = "finished"
```

PR #660, opened 2024-01-06
Allow passing a function to FasterWhisperPipeline.transcribe()

b0vik:
This'll enable custom progress bar implementations, e.g.
```python3
import whisperx
import gc 

device = "cuda" 
audio_file = "example.webm"
batch_size = 8 # reduce if low on GPU mem
compute_type = "int8" # change to "int8" if low on GPU mem (may reduce accuracy)

model = whisperx.load_model("large-v2", device, compute_type=compute_type)

def testpf(progress):
    print(f"Current progress as a fraction: {progress}")

audio = whisperx.load_audio(audio_file)
result = model.transcribe(audio, batch_size=batch_size, combined_progress=True, progress_function=testpf)
print(result["segments"]) # before alignment
```
```bash
$ python3 test.py                                                                                                                                                           
Detected language: en (1.00) in first 30s of audio...                                                                                                                    
Current progress as a fraction: 0.03333333333333333                                                                                                                      
Current progress as a fraction: 0.06666666666666667                                                                                                                      
Current progress as a fraction: 0.1                                                                                                                                      
Current progress as a fraction: 0.13333333333333333                                                                                                                      
Current progress as a fraction: 0.16666666666666666
```

### DIFF
--- a/whisperx/alignment.py
+++ b/whisperx/alignment.py
@@ -104,6 +104,7 @@ def align(
     return_char_alignments: bool = False,
     print_progress: bool = False,
     combined_progress: bool = False,
+    progress_function = None
 ) -> AlignedTranscriptionResult:
     """
     Align phoneme recognition predictions to known transcription.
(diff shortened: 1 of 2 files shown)


casic:
And  what about  alignment and diarization ??  Can add  function for
progress ??

На нд, 7.01.2024 г. в 0:29 b0vik ***@***.***> написа:

b0vik:
> And what about alignment and diarization ?? Can add function for progress ??  

Yeah, that's on me, sorry. I just added one for alignment. It doesn't look like `DiarizationPipeline` has an implementation for showing progress at the moment, so it'll be a bit less trivial to implement.

stex:
I also added a PR for this which passes `current` and `total` instead of the already computed value in #620

b0vik:
That's probably a better idea.

PR #675, opened 2024-01-25
fix package version

zacsun:


Terseus:
Please don't merge this, pinning the exact version for packages in a library is a bad idea, it prevents from using the library in an application with slightly different package versions, specially with `setuptools` I see no reason to pin it.

Swami-Abhinav:
Do not merge.

m-bain:
mob rules haha

Issue #674, opened 2024-01-24
Download in a directory option not available for DiarizaitonPipeline?


utility-aagrawal:
Hi,

I need to have all my models locally for my use case. I see that transcription has download_root parameter and align model has mode_dir parameter for this purpose. Is there no such parameter for DiarizationPipeline?

I can submit a PR for this change if you think this feature would be helpful. Let me know. I believe it will really useful for people who want to run everything locally. I understand that once model is downloaded from huggingface, it runs locally but I don't have an option to download in my target environment and need to have my models locally available beforehand.  Thanks!

utility-aagrawal:
@m-bain , Thanks for sharing your impressive work! Could you comment on this request please?

utility-aagrawal:
@m-bain , Could you comment on this please?

metheofanis:
Diarization is running locally anyway. It is only "gated" to hugging face. If you want to run it "ungated", see [this FAQ section in pyannote githab](https://github.com/pyannote/pyannote-audio/blob/develop/FAQ.md#can-i-use-gated-models-(and-pipelines)-offline)   
The second option reads: "[Can I use gated models (and pipelines) offline?](https://github.com/pyannote/pyannote-audio/blob/develop/FAQ.md#can-i-use-gated-models-(and-pipelines)-offline)"

This helped me do what you have described.

utility-aagrawal:
Thanks for your suggestion, @metheofanis! I'll check it out.

Barabazs:
This is now fixed in #1285

PR #691, opened 2024-02-06
Enable offline use for DiarizationPipeline and let users choose if they want to print suppress_numerals message.

utility-aagrawal:
Hi,

I have made a couple of changes that I think will be useful for others:

1) The first change is related to issue #674. Currently, Diarization pipeline doesn't have an option load model from a directory. I enabled that by following the tutorial here: "[Can I use gated models (and pipelines) offline?](https://github.com/pyannote/pyannote-audio/blob/develop/FAQ.md#can-i-use-gated-models-(and-pipelines)-offline)". In short, users can download the config.yaml and the pytorch model files from huggingface locally and use them in the pipeline.

2) Currently, when you enable suppress_numerals flag, a huge message is printed on the console. I introduced a flag so that users can choose if they want this print message or not.

Let me know if you have any questions. Thanks!

### DIFF
--- a/whisperx/asr.py
+++ b/whisperx/asr.py
@@ -171,7 +171,7 @@ class FasterWhisperPipeline(Pipeline):
         return final_iterator
 
     def transcribe(
-        self, audio: Union[str, np.ndarray], batch_size=None, num_workers=0, language=None, task=None, chunk_size=30, print_progress = False, combined_progress=False
+        self, audio: Union[str, np.ndarray], batch_size=None, num_workers=0, language=None, task=None, chunk_size=30, print_progress = False, combined_progress=False, supress_numerals_print_flag = True
     ) -> TranscriptionResult:
         if isinstance(audio, str):
             audio = load_audio(audio)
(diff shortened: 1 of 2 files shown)


GroovyDan:
@utility-aagrawal I believe you can already use the diarization pipeline offline by just specifying a local path to a `config.yaml` as the `model_name`:

diarize_model = whisperx.DiarizationPipeline(
    model_name=os.path.join(MODEL_DIR, DIARIZATION_FOLDER, "config.yaml"),
    device=DEVICE
)

rimeikis:
> @utility-aagrawal I believe you can already use the diarization pipeline offline by just specifying a local path to a `config.yaml` as the `model_name`:
> 
> diarize_model = whisperx.DiarizationPipeline( model_name=os.path.join(MODEL_DIR, DIARIZATION_FOLDER, "config.yaml"), device=DEVICE )

Could you please elaborate? I used [this guide](https://github.com/pyannote/pyannote-audio/blob/develop/tutorials/applying_a_pipeline.ipynb) to download the config.yml and pytorch_model.bin files, but when I provide it in the model_name argument, I get an error: 
`
huggingface_hub.utils._validators.HFValidationError: Repo id must be in the form 'repo_name' or 'namespace/repo_name'
`

GroovyDan:
@rimeikis What does your config.yml file look like? Have you updated it to point to the local pytorch_model.bin files you downloaded?

utility-aagrawal:
@GroovyDan , you are right! I was able to load diarization model from local using your suggestion. Closing this PR. Thank you!

digikar99:
> Have you updated it to point to the local pytorch_model.bin files you downloaded?

Could someone point out what exact change this is supposed to be?

digikar99:
I suppose the following will be the new config.yaml:

```
pipeline:
  name: pyannote.audio.pipelines.VoiceActivityDetection
  params:
    segmentation: pytorch_model.bin

params:
  min_duration_off: 0.09791355693027545
  min_duration_on: 0.05537587440407595
  offset: 0.4806866463041527
  onset: 0.8104268538848918
```

I do run into errors with this, but I will double check and create an issue if required.

PR #710, opened 2024-02-22
Updated faster-whisper version

debalabbas:
Installing library leading to conflicts due to version issue in fasterwhisper.
Faster whisper does not have branch `0.10.0` anymore. It has been updated to `v0.10.1`.

Have done the necessary changes in the requirements.txt.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 torch>=2
 torchaudio>=2
-faster-whisper @ git+https://github.com/SYSTRAN/faster-whisper.git@0.10.0
+faster-whisper @ git+https://github.com/SYSTRAN/faster-whisper.git@v0.10.1
 transformers
 pandas
 setuptools>=65


Terseus:
As I understand, when WhisperX started using faster whisper there wasn't a 0.10.0 release at the time, so it made sense to install from a branch.

That's no longer the case, we can find not only the version 0.10.1 in [pypi](https://pypi.org/project/faster-whisper/#history) but also the 1.0.0, released 3 hours ago.

Now that you're at it, @debalabbas can you please change the requirement to use just `faster-whisper` and let it install from pypi? What do you think about it @m-bain?

m-bain:
yeah makes sense right now im afk so can't verify the environment changes

Barabazs:
Thank you for your contribution, however this change is now obsolete due to a previously merged PR.

PR #712, opened 2024-02-22
Change faster-whisper dependency from `0.10.0` to `v0.10.0`

nealmanaktola:
pip install fails right now. 

I am guessing they renamed the tag?


nealmanaktola:
seems like a duplicate

Issue #711, opened 2024-02-22
Solved: TypeError: TranscriptionOptions.__new__() got an unexpected keyword argument 'max_new_tokens'


longdaothanh:
TypeError: TranscriptionOptions.__new__() got an unexpected keyword argument 'max_new_tokens'

Anyone has an idea how to fix this or has similar issues?

Problem Solved:
Change faster-whisper~=0.10.0 in requirements.txt to the newest version 1.0.0 then install -> working

longdaothanh:
![Screenshot 2024-02-22 154235](https://github.com/m-bain/whisperX/assets/115090741/78f90a10-34e7-48e0-ad12-8680167eb593)

longdaothanh:
![Screenshot 2024-02-22 190759](https://github.com/m-bain/whisperX/assets/115090741/9da3c2dc-62a2-4499-8647-aa3b51ba476c)

SobolevskyDmitry:
Seems new faster-whisper broke it
https://github.com/SYSTRAN/faster-whisper/releases

For me `faster-whisper-0.10.0` worked fine, but today I got the same error as you and checked my deps and found a new version: `faster-whisper-0.10.1`

SobolevskyDmitry:
I'm wondering would WhisperX work with https://github.com/SYSTRAN/faster-whisper/releases/tag/v1.0.0 ?

longdaothanh:
Hi, I just solved the problem 

> I'm wondering would WhisperX work with https://github.com/SYSTRAN/faster-whisper/releases/tag/v1.0.0 ?

Yes, I just tried and it worked perfect now.

msfuko:
Hi, how do you override the faster-whisper from `git+https://github.com/m-bain/whisperx.git@d6562c26da467ca49866a4173c0e723f3837f367`? or do you fork it?

davidmartinrius:
> Hi, how do you override the faster-whisper from `git+https://github.com/m-bain/whisperx.git@d6562c26da467ca49866a4173c0e723f3837f367`? or do you fork it?

after installing whisperx you can do this:

pip install --upgrade faster-whisper==1.0.0

sushk0841:
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
[<ipython-input-10-30a4dd61aaad>](https://localhost:8080/#) in <cell line: 1>()
----> 1 model = whisperx.load_model("large-v2", device, compute_type=compute_type)

[...]

PR #713, opened 2024-02-22
Update requirements.txt

jnuvvula9:
faster-whisper was just updated a couple of hours ago

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 torch>=2
 torchaudio>=2
-faster-whisper @ git+https://github.com/SYSTRAN/faster-whisper.git@0.10.0
+faster-whisper @ git+https://github.com/SYSTRAN/faster-whisper.git@1.0.0
 transformers
 pandas
 setuptools>=65


msfuko:
This PR also fixes this issue: https://github.com/m-bain/whisperX/issues/711

Terseus:
@jnuvvula9 Please, can you change the requirement to use pypi (`faster-whisper >= 1.0.0`) instead of installing from Github?

See https://github.com/m-bain/whisperX/pull/710#issuecomment-1959400388

davidmartinrius:
This update won't work. I already tested it.

deep-pipeline:
Just visiting the repo and looking at the 29 PRs and thinking that some are duplicates or overlapping PRs which have been overtaken and could be dropped with a bit of sanity checking to close off loose ends.

@davidmartinrius has this PR ( #713 ) been overtaken by any further updates and/or can you clarify if issue #711 is still live issue or not (i.e. can it closed - I've nudged the original poster). 

If so we could maybe suggest to repo owner that this PR request is closed or else a suitable fix is incorporated if one is needed? Thanks.

PR #716, opened 2024-02-22
fix: update faster-whisper dependencies

### LINKED ISSUE #715: Faster Whisper version update
Seems like faster whisper broke, and pip doesn't recognize v0.10.0 anymore. You can see it in the releases here.

https://github.com/SYSTRAN/faster-whisper

Can we change the requirement to not be versioned?

cococig:
Fix #715

As pointed out in #715, the current requirements.txt fails to install faster-whisper. Additionally, as mentioned in https://github.com/m-bain/whisperX/pull/710#issuecomment-1959400388, the PyPI version of faster-whisper has already been updated, and there is no longer a need to install directly from GitHub.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 torch>=2
 torchaudio>=2
-faster-whisper @ git+https://github.com/SYSTRAN/faster-whisper.git@0.10.0
+faster-whisper~=0.10.0
 transformers
 pandas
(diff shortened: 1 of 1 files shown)


aldrinjenson:
After some trial and error, I found that installing from repo now works!

`pip install git+https://github.com/m-bain/whisperx.git`

Issue #723, opened 2024-02-28
'speechbrain' must be installed to use 'speechbrain/spkrec-ecapa-voxceleb' embeddings


stepakuryaev:
I'm trying to install this project, I'm using PyCharm, a Python project, python 3.10.11, cuda 11.8, torch==2.0.0, torchvision==0.15.1, torchaudio==2.0.1, pyannote/speaker-diarization@2.1, and when, after installing all these components, I try to run the project, I get the following error: "'speechbrain' must be installed to use 'speechbrain/spkrec-ecapa-voxceleb' embeddings" and the program crashes. What could be the problem?

MelluChung:
I’ve encountered the same issue. 

I am able to import speechbrain directly within the environment, so the problem should not be related to speechbrain not being installed.

lukasbirki:
Happened to me as well, I am unable to import the package using Colab!

davidmartinrius:
Solution:
pip install -U speechbrain==0.5.16

The version 1.0.0 of speechbrain is not working well with whisperx.

vcombey:
It also broke the last version in replicate:
https://replicate.com/victor-upmeet/whisperx
when activating speaker diarization.
Do you know when it will be updated ?

thanks 🙏

PR #727, opened 2024-02-29
Speechbrain+faster whisper downgrade needed to make WhisperX work. 

davidmartinrius:
Hello!

This project only will work with speechbrain 0.5.16 and faster-whisper 0.10.1

Recently speechbrain upgraded to 1.0.0 which is [not compatible](https://github.com/m-bain/whisperX/issues/723) with the current version of whisperx.

On the other hand faster-whisper 1.0.0 has [breaking changes](https://github.com/m-bain/whisperX/issues/711) with the current latest version of whisperx too.

I already tested this in my project and made it work https://github.com/davidmartinrius/speech-dataset-generator

So, I just updated requirements.txt to make whisperX latest version work again.

Thank you!

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,8 @@
 torch>=2
 torchaudio>=2
-faster-whisper==1.0.0
+faster-whisper==0.10.1
 transformers
 pandas
 setuptools>=65
 nltk
+speechbrain==0.5.16


davidmartinrius:
Great! It seems it has been fixed. Thanks @MahmoudAshraf97

PR #739, opened 2024-03-11
Accept torch.Tensor as input

dkurt:
Useful to reduce data transfer between GPU and CPU to make `transcribe` when `audio` is already on GPU.

/cc @m-bain


Barabazs:
thanks for this. unfortunately this has been sitting for two years and the codebase looks pretty  different now. There might also a bug where passing a tensor actually crashes on the default Pyannote VAD path. Happy to review a fresh PR if you want to take another crack at it against current main.

PR #746, opened 2024-03-15
added new languages (example: bengali(bn),nepali etc)

redfieldC:
whisperX still lacks in laguages it can trasncribe, hence added more languages, please review, and correct the repositoriesit they are incorrect


m-bain:
we use torch ones for some of these languages. And you have to test a language before adding it

redfieldC:
Hi, please review the commit, I have tested the languages, it works

Issue #738, opened 2024-03-11
Verbose transcription doesn't work


DryHam:
Hi

In the orginal Openai-Whisper in Python you can set verbose=True, e.g.
result = model.transcribe(input_file,
                          language='en',
                          initial_prompt=prompt,
                          fp16=False,
                          verbose=True)

This prints the transcription output as the program runs so you can observe any errors in the transcription before it finalises, e.g.
[00:00.000 --> 00:07.000]  Okay.
[00:30.000 --> 00:43.680]  I'm making inquiries into a workplace dispute that
[00:43.680 --> 00:50.160]  has reportedly occurred on Saturday 6 January
[00:50.160 --> 00:51.160]  2024.

Judging from what I can see in transcribe.py, WhisperX has verbose=True set by default. However, nothing is printed while the program is running so you can't observe the transcription before it finalises. I have attempted to set print_progress = True in model.transcribe(), e.g.
result = model.transcribe(audio, batch_size=batch_size, print_progress=True)

However, this just prints out a percentage completion rate, not the transcription itself, e.g.
Progress: 1.43%...
Progress: 2.86%...
Progress: 4.29%...
Progress: 5.71%...
Progress: 7.14%...
Progress: 8.57%...

Is there a way to print out the transcription in Python as the program is running, or is this an error or an intended change with WhisperX?

PR #759, opened 2024-03-27
Add ultization to verbose flag

H4CK3Rabhi:
https://github.com/m-bain/whisperX/issues/738
had the issue with non-functionality of verbose flag and suggested handling it like openai-whisper

Quote from https://github.com/m-bain/whisperX/issues/738
```
In the orginal Openai-Whisper in Python you can set verbose=True, e.g.
result = model.transcribe(input_file,
language='en',
initial_prompt=prompt,
fp16=False,
verbose=True)

This prints the transcription output as the program runs so you can observe any errors in the transcription before it finalises, e.g.
[00:00.000 --> 00:07.000] Okay.
[00:30.000 --> 00:43.680] I'm making inquiries into a workplace dispute that
[00:43.680 --> 00:50.160] has reportedly occurred on Saturday 6 January
[00:50.160 --> 00:51.160] 2024.

Judging from what I can see in transcribe.py, WhisperX has verbose=True set by default. However, nothing is printed
```

Handled the same

### DIFF
--- a/whisperx/asr.py
+++ b/whisperx/asr.py
@@ -171,7 +171,7 @@ class FasterWhisperPipeline(Pipeline):
         return final_iterator
 
     def transcribe(
-        self, audio: Union[str, np.ndarray], batch_size=None, num_workers=0, language=None, task=None, chunk_size=30, print_progress = False, combined_progress=False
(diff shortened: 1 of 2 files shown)


Barabazs (approved):
Thank you for the contribution!

PR #776, opened 2024-04-11
Change the alignment model for Vietnamese language

nguyenvulebinh:
Since the current model is a wav2vec2 pre-trained model for Vietnamese audio, it won't work with alignment tasks. To make it work as expected, I recommend chaining to a fine-tuned ASR version.


Vanh253333:
I tried that fine-tuned ASR version you recommend, but it still doesn't work with alignment tasks.

PR #781, opened 2024-04-16
Update requirements.txt

remic33:
Update torchaudio to 2.2 to match pyannote update. 
Did not test it with other part f the project


Barabazs:
Why do you think this is needed? Is this solving a specific issue?

remic33:
> Why do you think this is needed? Is this solving a specific issue?

useless now, new pr made today

PR #786, opened 2024-04-19
Update README.md with instructions on how to install with pip

rodrigomeireles:
Current PyTorch instructions don't work out of the box because the `lit` package doesn't exist in the `cu118` index. Installing `lit` in a previous step solves the issue.

See https://github.com/pypa/pip/issues/12050#issuecomment-1948879030 for more details.


Barabazs:
`lit` isn't a direct dependency of whisperX though?

PR #814, opened 2024-05-27
Update faster-whisper to 1.0.2 to enable model distil-large-v3

moritzbrantner:


Barabazs (approved):
Thank you for your contribution

Issue #621, opened 2023-12-10
"--max_line_count" only works if "highlight_words" is True


DysnomiaDreams:
Everytime I set "max_line_count" it's ignored; unless I turn "highlight_words" to True, which kinda fixes it, aside from some weird glitching which can be seen in the gif.

With "highlight_words" as False:
`whisperx arcane.mp4 --model large-v3 --highlight_words False --max_line_width 42 --max_line_count 1 --output_format srt `
![vlcsnap-2023-12-10-11h14m49s395](https://github.com/m-bain/whisperX/assets/46659080/8df473e2-30d4-49be-9c56-47ee2092f0a0)

With "highlight_words" as True:
`whisperx arcane.mp4 --model large-v3 --highlight_words True --max_line_width 42 --max_line_count 1 --output_format srt` 
![whisperx-bug](https://github.com/m-bain/whisperX/assets/46659080/fb313e01-2a0f-435c-862b-e5d1e8d6e2ff)

Issue #608, opened 2023-11-28
Times problem with v3 model


carolinaxxxxx:
Hi,

It seems that when using the --max_line_width 42 --max_line_count 2 option there are incorrect timings in the output srt file:

Example:

```
1
00:00:15,912 --> 00:00:17,293
Good morning, loves! We're going to get 
breakfast! We're going to get breakfast at a new

2
00:00:19,935 --> 00:00:21,456
spot. It's called Spoon. Can't wait for you guys to 
see it. Enjoy! Please enjoy this Shed Eggs. Enjoy!

3
**00:01:02,962 --> 00:01:23,152**
hi guys oh hi sweethearts and welcome back to my 
channel welcome back to another video this is a

4
**00:01:02,962 --> 00:01:23,152**
highly anticipated video this is my birthday vlog 
thanks per welcome to my birthday vlog and it has

5
**00:01:02,962 --> 00:01:23,152**
been very eventful i'm not gonna lie am i too bright 
for you guys i probably am too bright let me

6
**00:01:02,962 --> 00:01:23,152**
probably like stand here welcome welcome welcome 
abroad welcome to thy sweet family welcome if

7
00:01:24,733 --> 00:01:46,482
you're new um if you're new to the channel i hope you 
like the video enough to give it a subscribe like

8
00:01:24,733 --> 00:01:46,482
down below and please comment guys i'm so happy i'm 
so sorry i need to calm down okay if you're new to the

9
00:01:47,842 --> 00:02:07,296
channel i hope you like the channel enough to give 
it a thumbs up and comment down below and leave a

10
00:01:47,842 --> 00:02:07,296
subscription to become part of this sweet family 
and become a sweetheart i hope that's more calm but

11
00:01:47,842 --> 00:02:07,296
yeah guys welcome to my birthday vlog we are about 
to okay wait i don't know what i'm saying about my
```

The times overlap, so the output file is poorly constructed and, as a result, read.

Tested on multiple audio materials with the v3 model.

carolinaxxxxx:
It looks like it's the same on the v2 model.

```
1
00:00:15,912 --> 00:00:17,293
Good morning, loves! We're gonna get breakfast! 
We're going to get breakfast at a new spot, it's

2
00:00:19,935 --> 00:00:22,217
called Spoon. Can't wait for you guys to see it. 
Enjoy! Please, gotta go back. I can't do a buy-out,

3
00:00:28,501 --> 00:00:31,403
[...]

PR #816, opened 2024-06-02
Fix to Timing Overlap Issue #608 and #621

FredBill1:
Currently the function `iterate_subtitles` does not compute the start and end time correctly for each combined subtitle segment. All it does is selecting the start and end time of the first segement in the combined segement, without considering the following segments or segment breaks. The start and end time of each word is ignored, causing the `--max_line_count` and `--max_line_width` to return segements with overlapped timing, as stated in issue #608 and #621.

https://github.com/m-bain/whisperX/blob/f2da2f858e99e4211fe4f64b5f2938b007827e17/whisperx/utils.py#L274

https://github.com/m-bain/whisperX/blob/f2da2f858e99e4211fe4f64b5f2938b007827e17/whisperx/utils.py#L281-L284

https://github.com/m-bain/whisperX/blob/f2da2f858e99e4211fe4f64b5f2938b007827e17/whisperx/utils.py#L316

---

I added a piece of code to compute the start and end time of the combined segment based on the time of each word in the segment. But I'm not sure how to utilize the `times` yielded by `iterate_subtitles`.

### DIFF
--- a/whisperx/utils.py
+++ b/whisperx/utils.py
@@ -278,15 +278,19 @@ class SubtitlesWriter(ResultWriter):
                 yield subtitle, times
 
         if "words" in result["segments"][0]:
-            for subtitle, _ in iterate_subtitles():
-                sstart, ssend, speaker = _[0]
+            for subtitle, times in iterate_subtitles():
+                # TODO: handle multiple segments with different start/end times and speakers
+                sstart, send, speaker = times[0]
+                has_timing = any(["start" in timing for timing in subtitle])
+                if has_timing:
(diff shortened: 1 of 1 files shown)


issaMbarki:
I edited the iterate_subtitles function and it gave me really good result now with ` max_line_count` and `max_line_width`, once I'll have sometime I'll create a pull request

tormki-uio:
This pr worked well for me, as we've been struggling with issues around max_line_width and max_line_count created duplicated timestamps for split segments

ghost:
why tf is this pr not merged yet?

Barabazs:
Closing in favor of #1347

Issue #785, opened 2024-04-19
Wrong cuda device (device_id > 0) on VAD in whisperx.load_model() method


grzegorz700:
When we use `whisperx.load_model(..., device='cuda', device_id=1)`, we get a quick peek at the wrong (cuda:0) GPU device instead of the desired one. The problem occurs due to different setups for the model and VAD (in the model, we split to device and device_id, but in VAD we use only  device name).

https://github.com/m-bain/whisperX/blob/f2da2f858e99e4211fe4f64b5f2938b007827e17/whisperx/asr.py#L347

### Bug visually:
<img width="696" alt="whisperX-wrong-gpu" src="https://github.com/m-bain/whisperX/assets/17555120/1ca28823-0e47-4f10-a712-ea84782edc01">

### Possible solution:
```python
if device == 'cuda' and device_id  and device_id> 0:
    device_vad = f"cuda:{device_id}"
else:
    device_vad = device
vad_model = load_vad_model(torch.device(device_vad), use_auth_token=None, **default_vad_options)
```

Barabazs:
Thank you for raising this issue and providing a good explanation! :pray:
Fixed in #835

PR #835, opened 2024-07-10
fix vad model load bug.

duj12:


3manifold (dismissed):
Can you provide more details on the reason behind the changes you submitted? If you don't set the device index, PyTorch will automatically use the default CUDA device, which is usually device 0. This is equivalent to: `device = torch.device('cuda:0')`.

createOne999:
https://github.com/m-bain/whisperX/issues/785#issue-2253389688

`
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:1 and cuda:0! (when checking argument for argument weight in method wrapper_CUDA__cudnn_batch_norm)
`


On multi-GPU setups, this may need to be set manually due to program requirements.

createOne999:
> Can you provide more details on the reason behind the changes you submitted? If you don't set the device index, PyTorch will automatically use the default CUDA device, which is usually device 0. This is equivalent to: `device = torch.device('cuda:0')`.

https://github.com/m-bain/whisperX/issues/785#issue-2253389688

`
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:1 and cuda:0! (when checking argument for argument weight in method wrapper_CUDA__cudnn_batch_norm)
`


On multi-GPU setups, this may need to be set manually due to program requirements.

Barabazs (approved):
Thank you

PR #862, opened 2024-08-20
transcribe: pass `suppress_numerals` into function to control per request

eschmidbauer:


Barabazs:
What's your reason for wanting to pass `suppress_numerals` to the function instead of setting it when you initialize the `FasterWhisperPipeline` class??

PR #867, opened 2024-08-25
Added local_files_only option on whisperx.load_model for offline mode

RoqueGio:
Adds the parameter local_files_only (default False for consistency) to whisperx.load_model so that the user can avoid downloading the file and return the path to the local cached file if it exists.

### DIFF
--- a/whisperx/asr.py
+++ b/whisperx/asr.py
@@ -267,6 +267,7 @@ def load_model(whisper_arch,
                model : Optional[WhisperModel] = None,
                task="transcribe",
                download_root=None,
+               local_files_only=False,
                threads=4):
     '''Load a Whisper model for inference.
     Args:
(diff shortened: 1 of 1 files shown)


madriss (commented):
Added local files only mode

whisperx/asr.py
```diff
@@ -277,6 +278,7 @@ def load_model(whisper_arch,
         language: str - The language of the model. (use English for now)
         model: Optional[WhisperModel] - The WhisperModel instance to use.
         download_root: Optional[str] - The root directory to download the model to.
+        local_files_only: Optional[bool] - If `True`, avoid downloading the file and return the path to the local cached file if it exists.
```
Barabazs, line 281:
```suggestion
        local_files_only: bool - If `True`, avoid downloading the file and return the path to the local cached file if it exists.
```
Optional[...] is for arguments that default to None, which is not the case here.

Barabazs (changes requested):
Just one minor change request please.

whisperx/asr.py
Barabazs: nvm, I didn't realise the PR was so old. I'll push the change

Barabazs (approved):

Barabazs:
Thank you for your contribution @RoqueGio !

PR #882, opened 2024-09-13
Updated asr.py to allow replacing initial_prompt on transcribe method.

clararoft:
Added the initial_prompt parameter to the transcribe method, allowing dynamic updates to the initial prompt set during model initialization.

PR #913, opened 2024-10-31
feat: Enable optional dynamic prompting for the FasterWhisperPipeline

jameshu88:
In this update to the asr.py file within the WhisperX repository, I've implemented an optional dynamic prompting feature for the FasterWhisperPipeline. This enhancement allows users to customize the prompts used during transcription dynamically, tailoring the input to better match specific contexts or tasks. By allowing dynamic prompts, users can provide context-specific cues that may lead to higher transcription accuracy, especially in specialized domains. This feature simplifies the process of refining transcription outputs without the need for extensive modifications to the underlying model or configuration, thus streamlining user workflows. 

I believe this improvement aligns with the goal of making WhisperX more adaptable and user-friendly, catering to various transcription scenarios and enhancing the overall performance of the FasterWhisperPipeline!

### DIFF
--- a/whisperx/asr.py
+++ b/whisperx/asr.py
@@ -67,7 +67,6 @@ class WhisperModel(faster_whisper.WhisperModel):
             res = []
             for tk in tokens:
                 res.append([token for token in tk if token < tokenizer.eot])
-            # text_tokens = [token for token in tokens if token < self.eot]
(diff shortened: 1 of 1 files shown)


Barabazs:
Duplicate of #882

PR #930, opened 2024-11-25
Update MANIFEST.in to include necessary files

frostming:


Barabazs:
Please add an explanation for why this is needed.

frostming:
To fix the sdist built from the source:

To reproduce, install `build` first: https://pypi.org/p/build , then run:

```
python -m build
```

You will notice a failure when building the wheel because by default `build` builds wheel from the result sdist.

Barabazs (approved):

Barabazs:
Thank you for clarifying and the contribution.

PR #947, opened 2024-12-17
Update Pyannote version + faster-whisper version 

remic33:
Pyannote 3.3.1 (required for numpy) 
faster-whisper 1.1.0


remic33:
Work now with readme update

PR #980, opened 2025-01-07
Divya/alignment speedup

divya-corpusant:
These are the necessary changes to perform backtracking in a cpu threadpool so we can get much better gpu utilization and overall throughput for the whisper alignment transform.
Has been tested by running our transform and spot checking outputs.
This PR also updates whisperx to the latest version, since one of the urls hosting some model was deprecated and it was broken as was

### DIFF
--- a/whisperx/alignment.py
+++ b/whisperx/alignment.py
@@ -3,12 +3,16 @@ Forced Alignment with Whisper
 C. Max Bain
 """
 from dataclasses import dataclass
-from typing import Iterable, Union, List
+from multiprocessing.pool import AsyncResult
+import time
+import traceback
+from typing import Iterable, Optional, Union, List
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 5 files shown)


whisperx/alignment.py
```diff
@@ -100,6 +104,185 @@ def load_align_model(language_code: str, device: str, model_name: Optional[str]
...
+                aligned_segments.append(aligned_seg)
+                continue
+
+            text_clean = "".join(segment["clean_char"])
+
+            print("getting trellis for sample", debug_id)
+            start_time_trellis = time.time()
+
+            trellis = get_trellis(emission, tokens, blank_id)
+            print("doing backtrack for sample", debug_id, ". Trellis took", time.time() - start_time_trellis)
```
divya-corpusant, line 168: Can clean this logging up a little bit

PR #982, opened 2025-01-08
Improved sentence segmentation based on word timestamps

Rumeysakeskin:
This PR improves sentence segmentation by handling long sentences and enforcing time-based breaks. Previously, all words between punctuation marks were grouped into a single segment, regardless of duration. This could lead to extremely long segments (60-80 seconds) which is problematic for many applications.

### DIFF
--- a/whisperx/alignment.py
+++ b/whisperx/alignment.py
@@ -67,14 +67,12 @@ DEFAULT_ALIGN_MODELS_HF = {
 
 def load_align_model(language_code: str, device: str, model_name: Optional[str] = None, model_dir=None):
     if model_name is None:
-        # use default model
         if language_code in DEFAULT_ALIGN_MODELS_TORCH:
             model_name = DEFAULT_ALIGN_MODELS_TORCH[language_code]
         elif language_code in DEFAULT_ALIGN_MODELS_HF:
             model_name = DEFAULT_ALIGN_MODELS_HF[language_code]
         else:
-            print(f"There is no default alignment model set for this language ({language_code}).\
-                Please find a wav2vec2.0 model finetuned on this language in https://huggingface.co/models, then pass the model name in --align_model [MODEL_NAME]")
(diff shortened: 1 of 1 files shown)


whisperx/alignment.py
```diff
@@ -89,17 +87,55 @@ def load_align_model(language_code: str, device: str, model_name: Optional[str]
             align_model = Wav2Vec2ForCTC.from_pretrained(model_name, cache_dir=model_dir)
         except Exception as e:
             print(e)
-            print(f"Error loading model from huggingface, check https://huggingface.co/models for finetuned wav2vec2.0 models")
-            raise ValueError(f'The chosen align_model "{model_name}" could not be found in huggingface (https://huggingface.co/models) or torchaudio (https://pytorch.org/audio/stable/pipelines.html#id14)')
```
m-bain, line 93: keep this

whisperx/alignment.py
```diff
@@ -89,17 +87,55 @@ def load_align_model(language_code: str, device: str, model_name: Optional[str]
...
+            prev_word = current_segment[-1]
+            if "end" not in prev_word or "start" not in word:
+                continue
+                
+            prev_word_end = prev_word["end"]
+            current_word_start = word["start"]
+            gap_duration = current_word_start - prev_word_end
+            
+            if gap_duration >= gap_threshold:
+                if current_segment:  # Boş segment eklemeyi önle
```
m-bain, line 126:
```suggestion
                if current_segment:  # Prevent adding empty segments
```

whisperx/alignment.py
```diff
@@ -89,17 +87,55 @@ def load_align_model(language_code: str, device: str, model_name: Optional[str]
...
+                    segments.append(current_segment)
+                current_segment = [word]
+            els
```
m-bain, line 129:
```suggestion
            else:
```

m-bain:
I feel like these changes should be in the subtitle writer

PR #995, opened 2025-01-13
fix vad_method is none

winking324:
vad_method is `None` by default, so `pyannote` should be chosen by default in this case?


Barabazs:
> vad_method is `None` by default, so `pyannote` should be chosen by default in this case?

@m-bain I see you've added this change in https://github.com/m-bain/whisperX/commit/aaddb83aa558ac3a4183d75748dc83c51e6253b1. Should we modify it so it defaults to pyannote if no vad_method was given?

m-bain:
yup, thanks @Barabazs. @winking324 i modiifed the PR to just default the function arg

PR #952, opened 2024-12-22
Speaker embeddings

agorman:
This is a simple change to allow the speaker embeddings to be added to the JSON output file. I don't often code in Python so I'm looking any feedback is welcome.

PR #997, opened 2025-01-14
Add option to return embeddings 

charnesp:
I added an option for returning embeddings, which can be very useful for speaker identification


charnesp:
Edit : the pull request https://github.com/m-bain/whisperX/pull/952 seems to do the same job.

PR #986, opened 2025-01-09
support timestamp for numbers.

bfs18:
updated get_trellis and backtrack to support align numbers.

```
import whisperx
import gc 

device = "cuda" 
audio_file = "--BhThOY2Ug_2.mp3"
batch_size = 16 # reduce if low on GPU mem
compute_type = "float16" # change to "int8" if low on GPU mem (may reduce accuracy)

# 1. Transcribe with original whisper (batched)
model = whisperx.load_model(
    "large-v3", device, compute_type=compute_type,
    asr_options={"word_timestamps": True, "without_timestamps": False})

# save model to local path (optional)
# model_dir = "/path/"
# model = whisperx.load_model("large-v2", device, compute_type=compute_type, download_root=model_dir)

audio = whisperx.load_audio(audio_file)
result = model.transcribe(audio, batch_size=batch_size)
print(result["segments"]) # before alignment

# delete model if low on GPU resources
# import gc; gc.collect(); torch.cuda.empty_cache(); del model

# 2. Align whisper output
model_a, metadata = whisperx.load_align_model(language_code=result["language"], device=device)
result = whisperx.align(result["segments"], model_a, metadata, audio, device, return_char_alignments=False)

print(result["segments"]) # after alignment
```

[--BhThOY2Ug_2.mp3](https://drive.google.com/file/d/1FfbDck2bmYpv1FyXxh1kxHU-GPeUY9E5/view?usp=sharing)

The output is 
```
Detected language: en (1.00) in first 30s of audio...
[{'text': " Thomas' uncle posted an emotional message on Facebook, and the link to a fundraiser to help cover medical costs has raised over $43,000. Matt Uris for CBS 11 News.", 'start': 0.031, 'end': 10.409}]
[...]

PR #999, opened 2025-01-15
Fix subtitle overlaps

lesca:
Hello,
This is a fix for subtitle overlaps that occurred after introducing [support for timestamps with numbers PR#986](https://github.com/m-bain/whisperX/pull/986).

In some cases, two subtitles could appear at the same time, as the figure below shows:

截屏2025-01-15 11 47 23

The last word is highlighted in the previous sentence (2nd line), and the first word is highlighted in the current sentence (1st line). 

To fix it, I add 0.03s to the start of the sentence and decrese 0.02s from the beginning of the sentence.

### DIFF
--- a/whisperx/alignment.py
+++ b/whisperx/alignment.py
@@ -306,6 +306,9 @@ def align(
         char_segments_arr = pd.DataFrame(char_segments_arr)
 
         aligned_subsegments = []
+        # sentecen offset in senconds to fix subtitle overlaps
+        sentence_start_offset = 0.03
+        sentence_end_offset = -0.02
         # assign sentence_idx to each character index
         char_segments_arr["sentence-idx"] = None
         for sdx2, (sstart, send) in enumerate(segment_data[sdx]["sentence_spans"]):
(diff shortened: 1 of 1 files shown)


Barabazs:
Correct me I'm wrong, but hard coding a time based offset does not seem like a proper solution that will work for all cases. This seems more like a workaround for a one-off case.

lesca:
Hello @Barabazs , yes it's a workaround. I am not capable to understand how exactly other parts works and fix it. I used it for myself and it works well. It's fine if the PR is rejected, and I hope there will be a good solution soon.

Barabazs:
@bfs18 have you encountered this issue when working on this feature? And could you implement a proper fix if needed?

bfs18:
Hi @lesca @Barabazs 

I’ve reviewed the feedback from [this issue](https://github.com/m-bain/whisperX/issues/1016) and believe the problem with leading and trailing timestamps can be resolved more effectively through the changes proposed in [this pull request](https://github.com/m-bain/whisperX/pull/1019).

The issue arises because the leading and trailing word boundary symbols may not always be present in the text during alignment. By explicitly adding these symbols, we can address the problem comprehensively.

@lesca Could you please test the proposed changes in your specific use case and share your feedback? Your input would be greatly appreciated to ensure the fix works as intended.

lesca:
@bfs18 I will test the PR#1019 and feedback within it.

Issue #1016, opened 2025-01-23
Alignment issues after #986 (support timestamp for numbers)


MarkMLCode:
It seems that the fix recently implemented in #986  (support timestamp for numbers) causes issues with the alignment of the last word in a segment. Whenever there is a sound at the end of the file, it seems that the entire space between the last word and the noise is now detected as the last word (about a second in my test). This even places the end of the word after the total duration of the file. In fact, I've noticed it doing this even when the file has no noise at the end (that is, it detects the last word a little after the end duration of the file).

File used : test.wav
[test sound file.zip](https://github.com/user-attachments/files/18524786/test.sound.file.zip)

File duration : 7.8
Last word detected in the list (before the change): {'word': 'back,', 'start': 6.903, 'end': 7.104, 'score': 1.0}
Last word detected in the list (after the change): {'word': 'back,', 'start': 6.941, 'end': 7.847, 'score': 0.966}

Transcribe function results : 
{'segments': [{'text': ' I focus my energy, aiming a forceful magic missile at the remaining spectres and try to send them reeling back,', 'start': 0.031, 'end': 7.827}], 'language': 'en'}

Align (before the change) : 

[...]

PR #1019, opened 2025-01-26
 make sure the leading and tailing word boundary exists.

bfs18:
Fix https://github.com/m-bain/whisperX/issues/1016
On the [wav2vec alignment tutorial page](https://pytorch.org/audio/stable/tutorials/forced_alignment_tutorial.html), word boundary symbols are present at the beginning and end. During the whisperX alignment process, the tailing word boundary might be missing, causing the tailing silence to be absorbed into the last word. Adding the tailing word boundary would resolve this issue.

### DIFF
--- a/whisperx/alignment.py
+++ b/whisperx/alignment.py
@@ -158,6 +158,13 @@ def align(
         else:
             per_word = text
 
+        # make sure the leading and tailing word boundary exists.
+        if model_lang not in LANGUAGES_WITHOUT_SPACES:
+            if not text.startswith(" "):
+                text = " " + text
+            if not text.endswith(" "):
+                text = text + " "
+
         clean_char, clean_cdx = [], []
         for cdx, char in enumerate(text):
             char_ = char.lower()


lesca:
@bfs18 I cloned the latest main branch (44e8bf5bb658ad6e1a80585b987fae2d48b9662c) and patched with this PR. From my recent transcriptions on serveral videos, this PR works well and I no longer see sublitles overlap. Thanks for your great works!

bfs18:
Hi @lesca I am glad to hear that. Thank you for your feedback.

bfs18:
Hi @Barabazs @m-bain 
I believe this fix works. It produces correct results for the test case mentioned in issue [#1016](https://github.com/m-bain/whisperX/issues/1016) (though the original poster has not yet confirmed). Additionally, it should resolve the related issue [#999](https://github.com/m-bain/whisperX/pull/999).

whisperx/alignment.py
```diff
@@ -158,6 +158,13 @@ def align(
         else:
             per_word = text
 
+        # make sure the leading and tailing word boundary exists.
+        if model_lang not in LANGUAGES_WITHOUT_SPACES:
+            if not text.startswith(" "):
+                text = " " + text
+            if not text.endswith(" "):
+                text = text + " "
```
Barabazs, line 166: Won't this result in double spaces?

whisperx/alignment.py
```diff
@@ -173,8 +180,8 @@ def align(
             elif char_ in model_dictionary.keys():
                 clean_char.append(char_)
                 clean_cdx.append(cdx)
-            else:
-                # add placeholder
+            elif char_ not in '!"\',.:;<=>?[\\]^_`{|}':
```
Barabazs, line 183: You mention the word boundary symbol from Wav2Vec2, but the page you linked only uses `|`. Can you explain why these other symbols are needed?

And imo an else-clause should be added as a fallback in case of unforeseen issues.

Barabazs (changes requested):
Thank you very much for looking into this! I added a few questions inline.

whisperx/alignment.py
bfs18: No, it won't. If there's already a leading or trailing space, no additional space will be added. So no need to worry about double spaces.

whisperx/alignment.py
bfs18: In the original code, symbols beyond "a-z" were ignored, and word boundaries were added to the string for alignment purposes. However, some of these ignored symbols (like numbers or the dollar sign $) may have pronunciations and require timestamps. To address this, a wildcard placeholder is needed to ensure these symbols are properly processed.

On the other hand, certain symbols (like !"',.:;<=>?[\]^_`{|}) can be absorbed into word boundaries without needing a wildcard. The condition in the code filters out these absorbable symbols and only adds wildcard placeholders for symbols that have explicit pronunciations.

As for the else-clause, it simply skips the symbol, so it’s not explicitly implemented in the current logic.

Let me know if you’d like further clarification or adjustments.

PR #1047, opened 2025-02-14
Allow usage of ffmpeg audio filters

acastin:
Allow users to use ffmpeg audio filters to clean up or normalize audio data, with a link to the official documentation and some examples

### DIFF
--- a/whisperx/audio.py
+++ b/whisperx/audio.py
@@ -22,7 +22,7 @@ FRAMES_PER_SECOND = exact_div(SAMPLE_RATE, HOP_LENGTH)  # 10ms per audio frame
 TOKENS_PER_SECOND = exact_div(SAMPLE_RATE, N_SAMPLES_PER_TOKEN)  # 20ms per audio token
 
 
-def load_audio(file: str, sr: int = SAMPLE_RATE) -> np.ndarray:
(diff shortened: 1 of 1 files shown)


Barabazs:
My 2 cents: this should be part of the users workflow, outside of the context of WhisperX.
1. it opens the door to shell/binary exploits
2. additional dependency headaches since ffmpeg isn't shipped with WhisperX
3. shell errors/exceptions are hard to handle and will result in a subpar experience

NielsMayer:
@acastin -- what kind of results and improvements are you seeing when employing, e.g., "EBU R 128" or "dynaudnorm" ? Is there a reason regular whisper doesn't employ this technique? Do any other variants?

```
ffmpeg_audio_filters: Optional[str]
        Apply ffmpeg audio filters (https://ffmpeg.org/ffmpeg-filters.html)
        "ebur128" to normalize loudness across the audio based on EBU R 128
        "dynaudnorm=p=0.5:s=5:g=15" to normalize volume dynamically over short windows
```

Thanks for pointing out this feature of ffmpeg... looks quite useful!

clort81:
As a rule: Stick to Unix Philosophy for maintainability -- Keep scope, bindings, dependencies limited.  Yes, people are forgetting this wisdom, which is why I write it again and again. 'App thinking' is a commerce model.  Not applicable in FOSS tool as here.
This is what scripting is for.  Tool pipelines in text, graphics, whatever. 
Btw 'normalize dynamically' is called compression in music engineering.

PR #1073, opened 2025-03-03
feat(hotwords): Pass hotwords option to faster-whisper

jademlc:
The hotwords argument exists but isn't passed to faster-whisper, this PR allows us to use hotwords


Barabazs (approved):
Thank you!

PR #1074, opened 2025-03-03
Basque alignment model

xezpeleta:
The following PR adds a phoneme-based alignment model for Basque language.

As a result we obtain high-precision word-level timestamps.

Demo: https://youtu.be/adB62l_PdYA

Thanks


Barabazs (approved):
Thank you!

jim60105:
@Barabazs We now have two 'eu.' Is this actually working?

xezpeleta:
Sorry, I didn't realize there was already an "eu" in the list. I'm going to add a new PR to fix it.

Barabazs:
> @Barabazs We now have two 'eu.' Is this actually working?

Thanks for catching this. Reverted in #1077

xezpeleta:
Ok, I've just tested it with two different alignment models, and I don't see any difference in alignment accuracy. So, I'm not going to suggest any changes to the current configuration.

Thanks, and sorry for the confusion!

PR #1098, opened 2025-04-05
feat: Add CUDA Toolkit 12.6 Installation Instructions and Update Dependencies uv.lock. Fix #1087

### LINKED ISSUE #1087: Its a mess. (Sorted Out)
Its a mess alright. 
install instructions does not work and causes all sort of incomplete or corrupt install where only parts of whisperX works or dont work at all because
stupid python libs.

Please sum up all needed deps to run this whisperX
I mean all of them, so i can freez the packages on the system and only get what is required.

jim60105:
Fix #1087

The primary modifications include adding the cu126 package repository source into the `pyproject.toml` file and the updating of `uv.lock` dependencies.

> [!IMPORTANT]  
> Please be aware that my testing has been conducted solely under Linux (container) environment. It is essential to verify it works on Windows before approval.

(I am a Fedora user and currently do not have access to a Windows machine with GPU card.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,15 @@ This repository provides fast automatic speech recognition (70x realtime with la
 
 <h2 align="left" id="setup">Setup ⚙️</h2>
 
+### 0. CUDA Installation
+
+To use WhisperX with GPU acceleration, install the CUDA toolkit 12.6 before WhisperX. Skip this step if using only the CPU.
+
+- For **Linux** users, install the CUDA toolkit 12.6 following this guide:
+  [CUDA Installation Guide for Linux](https://docs.nvidia.com/cuda/cuda-installation-guide-linux/).
+- For **Windows** users, download and install the CUDA toolkit 12.6:
+  [CUDA Downloads](https://developer.nvidia.com/cuda-12-6-3-download-archive).
+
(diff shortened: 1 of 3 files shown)


jim60105:
numpy v2.1.2 does not support Python 3.9
https://github.com/numpy/numpy/releases/tag/v2.1.2

The following is the numpy list provided by pytorch
https://download.pytorch.org/whl/numpy/

It can be seen that if 2.1.2 is not used, the previous version would be 1.26.3.
The original project requires `numpy>=2.0.2`. I'm not sure what would happen if we downgrade to version 1.26, but I think there's a greater chance of issues arising.

**Please decide whether to downgrade numpy or abandon support for Python 3.9.**

jim60105:
@Barabazs @m-bain Kindly take a moment to review this.

jim60105:
A recent commit cd59f21d1a0fe1269d153a1aba4925cdc24cc4aa downgraded ctranslate2 to version 4.4.0, meaning we will remain on cudnn8.
Since cu126 is compatible with cudnn9, this PR is no longer applicable. 

Additionally, as far as I know, stay on cudnn8 indicates that we do not support Blackwell (RTX 50XX), though I do not have the hardware to verify this.

PR #1101, opened 2025-04-08
feat: diarization model env config

bgdnvk:
Removing the hardcoded value from `diarization` so we can take it from the environment with `DIARIZATION_MODEL_CONFIG` but also keeping the original default `pyannote/speaker-diarization-3.1`


sonicfields (approved):

Barabazs (approved):
Thank you!

pintoseb:
Hello, I've receiving this error after this change, AttributeError: module 'whisperx' has no attribute 'DiarizationPipeline'
 could be related to? I'm running on google colab. Thks

PR #1110, opened 2025-04-11
feat: adds KBLab/kb-whisper-large as alignment model to Swedish

Npahlfer:
The national library of Sweden has released a model that is very efficient with the Swedish language and this PR adds it to the list of alignment models.


MahmoudAshraf97:
The alignment model needs to be a grapheme CTC model, whisper is an encoder-decoder model which is vastly different

Barabazs:
@Npahlfer It seems you're looking for this model "KBLab/wav2vec2-large-voxrex-swedish"

Npahlfer:
Thanks for the help! I changed the model to the CTC model.

Barabazs:

Issue #1038, opened 2025-02-09
ImportError: libctranslate2-d3638643.so.4.4.0: cannot enable executable stack as shared object requires: Invalid argument


GUUser91:
```
whisperx --model large-v3 --language en audio.flac
Traceback (most recent call last):
  File "/home/user/WhisperX/venv/bin/whisperx", line 5, in <module>
    from whisperx.transcribe import cli
  File "/home/user/WhisperX/venv/lib/python3.11/site-packages/whisperx/__init__.py", line 4, in <module>
    from .asr import load_model
  File "/home/user/WhisperX/venv/lib/python3.11/site-packages/whisperx/asr.py", line 5, in <module>
    import ctranslate2
  File "/home/user/WhisperX/venv/lib/python3.11/site-packages/ctranslate2/__init__.py", line 21, in <module>
    from ctranslate2._ext import (
ImportError: libctranslate2-d3638643.so.4.4.0: cannot enable executable stack as shared object requires: Invalid argument
```

I was able to fix this problem by downgrading to glibc-2.40-1 on my arch based artix linux distro.

```
sudo pacman -U glibc-2.40-1-x86_64.pkg.tar.zst
```
I found the package here.
https://archive.org/download/archlinux_pkg_glibc

The CTranslate2 page had the same issue
https://github.com/OpenNMT/CTranslate2/issues/1849#issuecomment-2639117406

If you uninstalled glibc and can't install the glibc package on arch or artix. Then you fix this problem by following this guide.
https://bbs.archlinux.org/viewtopic.php?id=142459

WasamiKirua:
thanks bro it was driving me crazy. i love linux but sometimes it sucks

chboishabba:
> [@WasamiKirua](https://github.com/WasamiKirua) couldn't write it better. Was about to throw my PC out of the window :D and fyi, I use [downgrade](https://aur.archlinux.org/packages/downgrade) to easily downgrade a package

I'm sure you're aware of  why downgrading is not recommended.

Hate9:
Yeah so, fair warning: I tried using `downgrade` to downgrade glibc, and it bricked my PC.

I know what I'm doing, and was confident I could fix it if this happened (which I did), but I would *not* recommend doing that if you aren't.

chboishabba:
@Hate9 that seems pretty unusual...

In any case wouldn't using the docker prevent this in future?

GUUser91:
[...]

PR #1126, opened 2025-05-01
fix: resolve dependency issues

Barabazs:
This has been tested on Google Colab with 3.10 & 3.11 and worked as it used to.

YMMV, but I also had to run `apt-get install libcudnn8 libcudnn8-dev` after installing whisperx.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,11 +9,11 @@ requires-python = ">=3.9, <3.13"
 license = { text = "BSD-2-Clause" }
 
 dependencies = [
-    "ctranslate2>=4.5.0",
+    "ctranslate2<4.5.0",
     "faster-whisper>=1.1.1",
     "nltk>=3.9.1",
     "numpy>=2.0.2",
-    "onnxruntime==1.19",
+    "onnxruntime>=1.19",
     "pandas>=2.2.3",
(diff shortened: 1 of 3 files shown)


uthark:
I have the same error as described in https://github.com/m-bain/whisperX/issues/1038

```
... redacted ...
    import whisperx.asr
  File "/app/.venv/lib/python3.10/site-packages/whisperx/asr.py", line 5, in <module>
    import ctranslate2
  File "/app/.venv/lib/python3.10/site-packages/ctranslate2/__init__.py", line 21, in <module>
    from ctranslate2._ext import (
ImportError: libctranslate2-d3638643.so.4.4.0: cannot enable executable stack as shared object requires: Invalid argument
```

but now workaround to bump ctranslate to the version 4.6.0 doesn't work due to this PR:

```
10.42 ERROR: Cannot install -r setup/requirements/requirements-forced-alignment.txt (line 14) and ctranslate2==4.6.0 because these package versions have conflicting dependencies.
10.42
10.42 The conflict is caused by:
10.42     The user requested ctranslate2==4.6.0
10.42     whisperx 3.4.2 depends on ctranslate2<4.5.0
```

what was the actual issue with blocking new version of `ctranslate2`? 
How can I fix the issue?

PR #1128, opened 2025-05-02
feat: improve CLI loading speed by lazy loading public API

Barabazs:
Changes introduced by this PR:
* refactored CLI entry point from (transcribe.cli to __main__.cli)
* refactored imports for lazy loading

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,25 +22,19 @@
   </a>      
 </p>
 
-
 <img width="1216" align="center" alt="whisperx-arch" src="https://raw.githubusercontent.com/m-bain/whisperX/refs/heads/main/figures/pipeline.png">
-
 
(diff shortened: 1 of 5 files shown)


Barabazs:
I'm all out of GPU credits for now, so can't test the latest commit for a while.

falahat:
@Barabazs I noticed that this PR removes the DiarizationPipeline from the public interface. However, some of the examples still reference this. 

Is there another way to perform Diarization without the DiarizationPipeline? As I understand it, we need this to create the diarization dataframe

Barabazs:
> @Barabazs I noticed that this PR removes the DiarizationPipeline from the public interface. However, some of the examples still reference this.
> 
> Is there another way to perform Diarization without the DiarizationPipeline? As I understand it, we need this to create the diarization dataframe

Hi @falahat which examples are still using it from the public API?
iirc they were updated to call DiarizationPipeline directly from whisperx.diarize

PR #1133, opened 2025-05-04
chore: restrict torch version to below 2.4 in dependencies (fix the libcudnn dependencies common issue)

jim60105:
A recent commit cd59f21d1a0fe1269d153a1aba4925cdc24cc4aa downgraded ctranslate2 to version 4.4.0, meaning we will remain on cudnn8.

torch depends on libcudnn9 from version 2.4.0 onward.
If we restrict torch<2.4.0, there is no need to manually install libcudnn8 and also save about 1GB disk space.

This PR:

- Update torch dependency to be below version 2.4.0 instead of at least 2.5.1
- Change torchaudio dependency to have no minimum version specified
- restrict onnxruntime to version 1.19 for python 3.9 compatibility

Related #1087, #1158

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,25 +97,6 @@ uv sync --all-extras --dev
 
 You may also need to install ffmpeg, rust etc. Follow openAI instructions here https://github.com/openai/whisper#setup.
 
-### Common Issues & Troubleshooting 🔧
-
-#### libcudnn Dependencies (GPU Users)
-
-If you're using WhisperX with GPU support and encounter errors like:
-
-- `Could not load library libcudnn_ops_infer.so.8`
-- `Unable to load any of {libcudnn_cnn.so.9.1.0, libcudnn_cnn.so.9.1, libcudnn_cnn.so.9, libcudnn_cnn.so}`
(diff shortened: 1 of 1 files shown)


jim60105:
@Barabazs This PR addresses the issue you mentioned in commit 6fe0a8784a22301e92624691193c26aab96e0c5e

jim60105:
Force-pushed to rebase onto the main branch.

jim60105:
After running `uv lock --upgrade` today, I received this build error.

[Build fails due to torch vulnerability CVE-2025-32434 requiring PyTorch v2.6+ upgrade](https://github.com/jim60105/docker-whisperX/issues/82)
> This issue was opened in my own repo because it comes from my own script.  
> However the content of the script is excerpted from the WhisperX project, so it should also occur here.

After carefully reading the CVE, I believe upgrading to torch >= 2.6.0 has become a MUST, so I will now abandon this PR.
The current version on the main branch is torch 2.5.1; I will provide another PR for torch >= 2.6.0 separately.

In addition, the limitation I mentioned in PR #1098 is likely to be encountered again. I will double-check the version dependencies in detail.

PR #1134, opened 2025-05-04
docs: document Docker image usage in README

jim60105:
- Add a new section to the README describing how to use pre-built Docker images for WhisperX with example commands.
- Provide a link to the Docker image repository for available tags.


coezbek:
Your Dockerfiles are not very easy to comprehend. Using statically compiled ffmpeg might be a security risk for some users. Why not use slightly larger base images which have proper package manager to load ffmpeg?

jim60105:
I designed my Docker project to be suitable for direct use with 'docker run', so users don't need to understand the Dockerfile.

If it were simple enough, I would have submitted the Dockerfile to this repository, but it's too complex to serve as a sample Dockerfile. 😅

Barabazs:
Hi @jim60105!

  Thanks for taking the time to create this PR and for building a Docker image for WhisperX.

  However, I think it's best not to list third-party packages and Docker images directly in the main README. The README should focus on the official installation and usage methods to
  avoid:
  - Maintenance burden of keeping external links up to date
  - Potential confusion about what's officially supported vs. community contributions
  - The challenge of deciding which community projects to include (there may be others in the future)

  **Alternative suggestion:** Would you be open to starting a GitHub Discussion instead? You could create a discussion in the "Show and tell" or "General" category where you share your Docker image with the community. This would:
  - Give your work visibility to users who need Docker solutions
  - Allow for ongoing community feedback and discussion
  - Keep the README focused on official documentation
  - Create a space where others can share similar tools/packages

  Thanks again for your contribution to the WhisperX ecosystem! 🙏

jim60105:
ok

Issue #1048, opened 2025-02-15
tensors used as indices must be long, int, byte or bool tensors


5arer:
Hello hello, thanks for the great work!

I have successfully been running this on many audios, but today I've hit one that resulted in error when doing alignment.
I'm using docker version which packages the binary but I think it shouldn't be relevant.
Any idea what this could be about?

Thanks!

```
>>Performing alignment...
Traceback (most recent call last):
  File "/home/1001/.local/bin/whisperx", line 8, in <module>
    sys.exit(cli())
             ^^^^^
  File "/home/1001/.local/lib/python3.11/site-packages/whisperx/transcribe.py", line 221, in cli
    result: AlignedTranscriptionResult = align(
                                         ^^^^^^
  File "/home/1001/.local/lib/python3.11/site-packages/whisperx/alignment.py", line 266, in align
    trellis = get_trellis(emission, tokens, blank_id)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/1001/.local/lib/python3.11/site-packages/whisperx/alignment.py", line 401, in get_trellis
    trellis[t, :-1] + get_wildcard_emission(emission[t], tokens[1:], blank_id),
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/1001/.local/lib/python3.11/site-packages/whisperx/alignment.py", line 426, in get_wildcard_emission
    regular_scores = frame_emission[tokens.clamp(min=0)]  # clamp to avoid -1 index
                     ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^
IndexError: tensors used as indices must be long, int, byte or bool tensors
Error executing transcription: Command '['docker', 'run', '-it', '-v', '.:/app', '-v', 'whisper_cache:/.cache', 'ghcr.io/jim60105/whisperx:no_model', '--', '--model', 'small', '--language', 'es', '--output_format', 'srt', '--compute_type', 'int8', 'audio.mp3']' returned non-zero exit status 1.

```

SyntevoAlex:
The faulty code was last changed in commit 65b2332e (2025-01-09)

SyntevoAlex:
I tried latest release v3.3.1 and it doesn't have this problem.

5arer:
[...]

PR #1146, opened 2025-05-15
Fix: Ensure integer tensor indexing in get_wildcard_emission() to avoid IndexError

HowardWhile:
**Bug Fix: IndexError in `alignment.py` when running alignment**

**Problem**
Sometimes when running alignment with WhisperX, I got this error:
```shell
IndexError: tensors used as indices must be long, int, byte or bool tensors
```

This happens in `alignment.py` at line 427:
```python
regular_scores = frame_emission[tokens.clamp(min=0)]
```
Looks like tokens can sometimes be a float tensor, but PyTorch only accepts integer types for indexing.

**How I fixed it**
I just added `.long()` to make sure it’s using integer indices:
```python
regular_scores = frame_emission[tokens.clamp(min=0).long()]
```
Pretty small change, but it fixes the crash.

**Notes**
- This bug is **intermittent** — doesn’t happen all the time.
- I ran into it while transcribing a Chinese audio file using `large-v2`.
- After the fix, the alignment step runs fine without errors.

### DIFF
--- a/whisperx/alignment.py
+++ b/whisperx/alignment.py
@@ -424,7 +424,7 @@ def get_wildcard_emission(frame_emission, tokens, blank_id):
     wildcard_mask = (tokens == -1)
 
     # Get scores for non-wildcard positions
-    regular_scores = frame_emission[tokens.clamp(min=0)]  # clamp to avoid -1 index
+    regular_scores = frame_emission[tokens.clamp(min=0).long()]  # clamp to avoid -1 index
 
     # Create a mask and compute the maximum value without modifying frame_emission
     max_valid_score = frame_emission.clone()   # Create a copy


tzvc:
Thanks! This is blocking our prod

micahjon:
FYI that a slightly different fix for the same issue was proposed here: https://github.com/m-bain/whisperX/issues/1048#issuecomment-2849500567

I don't have the context to know the best solution, but just wanted to connect this PR to the open issue.

tzvc:
@m-bain Any thoughts on this? This is quite critical

Barabazs (approved):
Thank you!

muzahidulislamhadi:
I had to fix this bug on the whisperx backend i run, months ago...

PR #1167, opened 2025-06-07
allow passing in num_speakers via cli

dehli:
I'm running into issues with diarization when using both `min_speakers` and `max_speakers` as the same value (2). I'm currently triggering the pipeline via the CLI and it would be great if I could similarly pass in `num_speakers` to be able to compare how that performs. 

I didn't test this change as I haven't gotten this project setup on my local machine, but with it being such a small change I figured I'd go ahead and submit the PR. Please let me know if you'd like any changes.


Barabazs:
> I didn't test this change

Please reopen this PR when you found the time to test your submission.

Issue #1173, opened 2025-06-15
IndexError: tensors used as indices must be long, int, byte or bool tensors


eropsakh:
'IndexError: tensors used as indices must be long, int, byte or bool tensors' occurred.

I changed line 427, whisperx/alignment.py
regular_scores = frame_emission[tokens.clamp(min=0)]
to
regular_scores = frame_emission[tokens.clamp(min=0).long()]

Then it worked well.

justadri:
this appears to be a duplicate of #1048, alternate solution in [comment](/m-bain/whisperX/issues/1048#issuecomment-2849500567)

PR #1181, opened 2025-06-26
Fix Tensor IndexError with Type Casting in alignment.py

nclarke0:
### What
Addresses an `IndexError` in `get_wildcard_emission` (alignment.py) when
`tokens` is not an integer dtype.

### Why
Only integer-typed tensors (`LongTensor`, etc.) can be used for indexing.
Without this fix, word-level alignment will crash on float32 token tensors.

### How
In `alignment.py`, before indexing, do:
  ```python
  clamped_tokens = tokens.clamp(min=0).long()
  regular_scores = frame_emission[clamped_tokens]
  ```
  
Closes m-bain/whisperX#1048
Closes m-bain/whisperX#1173

### DIFF
--- a/whisperx/alignment.py
+++ b/whisperx/alignment.py
@@ -424,7 +424,8 @@ def get_wildcard_emission(frame_emission, tokens, blank_id):
     wildcard_mask = (tokens == -1)
 
     # Get scores for non-wildcard positions
-    regular_scores = frame_emission[tokens.clamp(min=0)]  # clamp to avoid -1 index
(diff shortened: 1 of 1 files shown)


micahjon:
Same issue as https://github.com/m-bain/whisperX/pull/1146

Tagging @Barabazs  on this, would be awesome to get a fixed merged soon. Thanks!

Barabazs:
Thank you for the contribution. I'll close this in favor of the earlier fix.

PR #1183, opened 2025-07-04
fix: reorder memory deallocation sequence

SHi-ON:
This PR merely suggests the proper sequence for releasing memory occupied by the loaded model and `torch.cuda` cache, promoting more effective resource cleanup in the provided Python example.


Barabazs:
Thanks for the contribution! The current order in the README is actually more efficient for memory cleanup.

PR #1205, opened 2025-08-05
Changed __init__.py to allow better hinting

JoFrhwld:
As `__init__.py` is currently written, the functions available directly from `whisper` don't allow for hinting to reach through directly to the typing and doc strings:

Screenshot 2025-08-05 at 11 14 16 AM

This rewritten version enables access to typing and docstrings:

Screenshot 2025-08-05 at 11 16 17 AM

### DIFF
--- a/whisperx/__init__.py
+++ b/whisperx/__init__.py
@@ -1,31 +1,12 @@
-import importlib
+from whisperx.alignment import load_align_model, align
+from whisperx.asr import load_model
+from whisperx.audio import load_audio
+from whisperx.diarize import assign_word_speakers
 
-
-def _lazy_import(name):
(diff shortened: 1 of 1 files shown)


falahat:
+1 to improving the type hints and exported interface.

However, I believe that they want to lazily import the modules, and the current PR eagerly imports them. 

In #1229 , I added a PR, but kept the lazy importing (using the getattr method)

Also, we can add an "if TYPE_CHECKING" guard behind the module imports so that we still import/eagerly export the types for type checkers while still doing lazy loading during actual runtime.

Barabazs:
Thanks, but we do need the lazy imports. So just replacing it with eager imports is not helpful.

PR #1221, opened 2025-09-02
optionally use temp files on load_audio() to avoid running out of RAM when dealing with long audio files

to-audiobook:
Very long audio files might crash whisperX during the call to `load_audio()` in case the system runs out of RAM.

This PR adds the parameter `useTmpFiles` to `load_audio()`, which makes `ffmpeg` resample the audio using a temporary file, instead of trying to do it all on memory, thus substantially increasing the audio length whisperX can handle.

NOTE: if the audio file is long enough to crash the system during the `load_audio()` call, requiring the usage of temporary files, the system will probably run out of memory during the diarization stage too. I dealt with that by splitting the source audio in two or more parts, using the alignment stage result timestamps to avoid splitting the audio in the middle of a sentence. That code is not including in this patch and, even if it was, it still has some issues because, if we split the audio, the diarization might assign different speaker IDs to the same speaker on each one of those parts.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -169,3 +169,5 @@ cython_debug/
 
 # PyPI configuration file
 .pypirc
+
+ssh/
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,7 +9,7 @@ requires-python = ">=3.9, <3.13"
 license = { text = "BSD-2-Clause" }
 
 dependencies = [
-    "ctranslate2<4.5.0",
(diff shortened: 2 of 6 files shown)


Barabazs:
Thanks for the PR! The dependency and API changes have all been addressed separately. 

Closing this one out, but the temp-file idea for load_audio() might we worth tracking. I'll open a separate issue for it.

PR #1229, opened 2025-09-09
Use PEP 562-style lazy exports while still allowing for easy type checking

falahat:
Currently, the __init__.py defines some methods with `args*` and `kwargs**` which intercept method calls and lazily load the real methods (i.e. `alignment.load_alignment_model`.) This is good for perf, but the exported API is tough to use and opaque. It was unclear to me what the input parameters should be and static type checkers were unhappy.

This PR changes the __init__.py to use PEP 562-style exports, using the getattr method on the module itself to lazily load modules.

However, it also includes an "if TYPE_CHECKING" clause. This explicitly exports the actual methods, types, etc for the static type checking. However, during runtime, these modules are still lazily loaded efficiently.

Also, this PR exposes the DiarizationPipeline class since we need the diarization dataframe (which this pipeline creates) in order to use the "assign_speakers" method anyway. Let me know if that was removed on purpose or is no longer supported.

I couldn't find any unit tests. I will manually test this with my project (I am working on transcribing and diarizing some audio data) and report findings here. So far, I've tested my branch and the imports / type checks work correctly as expected.

Let me know if there is any specific testing I should perform other than basic audio transcription


falahat:
Also, I apologize for the numerous commits, I recommend a "squash" merge if you approve the PR

falahat:
Note, this PR also improves the type checking in diarizer.py to clarify the two (technically 3) ways we can call the diarize model and how it can return the speaker_embeddings or npt

falahat:
I created an issue, #1232 to track this

PR #1285, opened 2025-11-07
Fix: Respect --model_dir and --model_cache_only during alignment

MrPrayer:
## **Problem:**  
During alignment, WhisperX ignores the `--model_dir` and `--model_cache_only` command-line arguments, causing:

- Alignment models to be cached in the default directory instead of the provided `--model_dir`
- Transformers to attempt downloads even when the model is already cached, resulting in unnecessary delays on disconnected system

## **Solution:**  
- Add `model_cache_only` support to `load_align_model()` so it correctly reflects the `--model_cache_only` argument  
- Pass `local_files_only=model_cache_only` to `Wav2Vec2Processor` and `Wav2Vec2ForCTC`  
- Pass values from `--model_dir` and `--model_cache_only` into the alignment model loader

## **Impact:**  
- Alignment now (partially) respects user caching settings  
- Eliminates unnecessary download attempts and long wait times when the model is already cached

## **Testing:**  
- Ran from command line using `--model_dir .models` and `--model_cache_only True`  
- Confirmed correct behavior: alignment model loaded from `.models` with no download attempts

**Disclaimer**: i do not code in python, i do not use github, i just jumped through the stack on errors and warnings, made changes that made sense to me, and verified that result fixes my problems. So this change requires proper validation in both logic and code syntax. 
There's also a call to bundle.get_model() earlier in load_align_model's code (for torchaudio models) that would probably benefit from model_cache_only handling, if torchaudio supports it. But i don't use them and didn't fiddle with the code in there.

### DIFF
--- a/whisperx/diarize.py
+++ b/whisperx/diarize.py
@@ -94,12 +94,13 @@ class DiarizationPipeline:
         model_name=None,
         token=None,
         device: Optional[Union[str, torch.device]] = "cpu",
+        cache_dir=None,
     ):
         if isinstance(device, str):
             device = torch.device(device)
         model_config = model_name or "pyannote/speaker-diarization-community-1"
         logger.info(f"Loading diarization model: {model_config}")
-        self.model = Pipeline.from_pretrained(model_config, token=token).to(device)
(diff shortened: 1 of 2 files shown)


Barabazs (approved):

PR #1313, opened 2025-12-26
fix: pytorch fixes for pytorch>2.6 blackwell

0xrushi:
Resolves errors like 

    raise pickle.UnpicklingError(_get_wo_message(str(e))) from None
_pickle.UnpicklingError: Weights only load failed. This file can still be loaded, to do so you have two options, do those steps only if you trust the source of the checkpoint.
	(1) In PyTorch 2.6, we changed the default value of the `weights_only` argument in `torch.load` from `False` to `True`. Re-running `torch.load` with `weights_only` set to `False` will likely succeed, but it can result in arbitrary code execution. Do it only if you got the file from a trusted source.
	(2) Alternatively, to load with `weights_only=True` please check the recommended steps in the following error message.
	WeightsUnpickler error: Unsupported global: GLOBAL pyannote.audio.core.task.Resolution was not an allowed global by default. Please use `torch.serialization.add_safe_globals([pyannote.audio.core.task.Resolution])` or the `torch.serialization.safe_globals([pyannote.audio.core.task.Resolution])` context manager to allowlist this global if you trust this class/function.

### DIFF
--- a/whisperx/vads/pyannote.py
+++ b/whisperx/vads/pyannote.py
@@ -5,11 +5,19 @@ from typing import Optional
 import numpy as np
 import torch
 from pyannote.audio import Model
+from pyannote.audio.core.model import Introspection
+from pyannote.audio.core.task import Problem, Resolution, Specifications
 from pyannote.audio.core.io import AudioFile
 from pyannote.audio.pipelines import VoiceActivityDetection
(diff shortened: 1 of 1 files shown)


RakeshReddyKondeti:
Hi @Barabazs 
This PR worked for me and resolved my issues. Please consider merging it. Thanks!

PR #1327, opened 2026-01-27
 fix: pin huggingface-hub<1.0.0 for pyannote-audio compatibility

Barabazs:
This PR pins huggingface-hub<1.0.0 to prevent the use_auth_token removal. Merely updating the arguments as seen in other PR's is not enough because the dependencies are also incompatible.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,7 +2,7 @@
 urls = { repository = "https://github.com/m-bain/whisperx" }
 authors = [{ name = "Max Bain" }]
 name = "whisperx"
-version = "3.7.4"
+version = "3.7.5"
 description = "Time-Accurate Automatic Speech Recognition using Whisper."
 readme = "README.md"
 requires-python = ">=3.9, <3.14"
--- a/uv.lock
+++ b/uv.lock
@@ -4016,7 +4016,7 @@ wheels = [
 
 [[package]]
 name = "whisperx"
-version = "3.7.4"
+version = "3.7.5"
 source = { editable = "." }
 dependencies = [
     { name = "av" },


bobsummerwill:
Greetings!

RE: "Did you actually try running the code with these changes?"

Yes, I've been running successfully with the change since November, and have transcribed numerous podcasts using WhisperX-3.7.4 and pyannote-audio-4.0.1:

https://github.com/strato-net/strato-transcripts/blob/3bd20929082bbe0f4311a973b64e03a449201e7b/scripts/install_packages_and_venv.sh#L385

I also need a SpeechBrain patch applied which you can see in the link above, but they have actually already made the same change in their develop branch, but have not released it yet.

Here are all the workarounds I have stacked:
https://github.com/strato-net/strato-transcripts/blob/main/WORKAROUNDS.md

Barabazs:
> Greetings!
> 
> RE: "Did you actually try running the code with these changes?"
> 
> Yes, I've been running successfully with the change since November, and have transcribed numerous podcasts using WhisperX-3.7.4 and pyannote-audio-4.0.1:
> 
> https://github.com/strato-net/strato-transcripts/blob/3bd20929082bbe0f4311a973b64e03a449201e7b/scripts/install_packages_and_venv.sh#L385
> 
> I also need a SpeechBrain patch applied which you can see in the link above, but they have actually already made the same change in their develop branch, but have not released it yet.
> 
> Here are all the workarounds I have stacked: https://github.com/strato-net/strato-transcripts/blob/main/WORKAROUNDS.md

Hi Bob, thanks for replying.

The reason I'm asking is because these changes in isolation were not compatible with the dependencies, hence the temporary workaround by pinning an older version of huggingface-hub.

Thanks for sharing your workaround doc. I'll refer to it once I can dedicate more time to properly test pyannote 4.

bobsummerwill:
You're welcome.

It's a little bit of a tangle, I know!

Best wishes and thanks for the follow-up.

PR #1329, opened 2026-01-27
Pass model_dir when initialising diarization model

philmcmahon:
The `Pipeline.from_pretrained` function call supports passing through cache_dir: https://github.com/pyannote/pyannote-audio/blob/6328b97b9489e1b6ae99f586df01c1c6fcac8b82/src/pyannote/audio/core/pipeline.py#L142

I think it makes sense to pass through model_dir so that it's consistent for diarization models as well as transcription.

I initially made this change in order to ensure that diarization works offline if the models have already been cached, but discovered that setting HF_HUB_OFFLINE (https://huggingface.co/docs/huggingface_hub/en/package_reference/environment_variables#hfhuboffline) achieves that. 

I tested this locally and verified transcription still works

### DIFF
--- a/whisperx/diarize.py
+++ b/whisperx/diarize.py
@@ -17,12 +17,13 @@ class DiarizationPipeline:
         model_name=None,
         use_auth_token=None,
         device: Optional[Union[str, torch.device]] = "cpu",
+        model_dir: Optional[str] = None,
     ):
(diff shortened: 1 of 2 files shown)


Barabazs:
Closing in favor of previously merged #1285

Issue #1318, opened 2026-01-05
[FEATURE] multiple output formats (`--output_format`) with CLI


ulmemxpoc:
# details

- `--output_format`  only allows one of [`["all", "srt", "vtt", "txt", "tsv", "json", "aud"]`](https://github.com/m-bain/whisperX/blob/d32ec3e3012ec4c0934f4088424c32f3f038b249/whisperx/__main__.py#L25)
- I want to choose two or more output formats, but not all (for example, `--output_format srt,txt`)

# potential implementation plan

1. update [`__main__.py`](https://github.com/m-bain/whisperX/blob/d32ec3e3012ec4c0934f4088424c32f3f038b249/whisperx/__main__.py#L25) to allow comma-separated values (for example, `--output_format srt,txt,json`)
2. update [`get_writer()` in `utils.py`](https://github.com/m-bain/whisperX/blob/d32ec3e3012ec4c0934f4088424c32f3f038b249/whisperx/utils.py#L448-L459) to loop through all `output_format`s

dugd:
I'd like to work on this, can I start?

PR #1361, opened 2026-02-25
feat: support multiple comma-separated output formats

dugd:
This PR adds support for specifying multiple output formats simultaneously using a comma-separated list in the `--output_format` CLI argument.

Now, users can selectively choose two or more formats (e.g., `srt,vtt,json`) without generating unnecessary formats. Backward compatibility with the behavior of the `all` value is preserved.

**How it works:**
`whisperx audio.mp3 --output_format srt,txt,json`
*Will generate only `srt`,`txt`,`json` files.*

closes: #1318

### DIFF
--- a/whisperx/__main__.py
+++ b/whisperx/__main__.py
@@ -6,7 +6,7 @@ import torch
 
 from whisperx.utils import (
     LANGUAGES,
-    OUTPUT_FORMATS,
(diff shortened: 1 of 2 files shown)


dugd:
Updated as requested.
Crucially, `--output_format` now defaults to `standard` (equivalent to the previous `all` behavior).

PR #1371, opened 2026-03-11
feat: add progress_callback to transcribe, align, and diarize

Barabazs:
## Summary

- Add `progress_callback` parameter to `transcribe()`, `align()`, and `DiarizationPipeline.__call__()` for real-time progress tracking (0-100%)
- Callback is independent to `print_progress`. Both can fire independently
- Diarization wraps the callback into pyannote's internal hook protocol, mapping segmentation (0-50%) and embedding (50-99%) steps into monotonic sub-ranges with 100% reserved for actual completion

Builds on the work started in previous PR's, which were a good start, but had several gaps that we needed to address.

## Usage example

```python
import whisperx
from whisperx.diarize import DiarizationPipeline

device = "cuda"
audio = whisperx.load_audio("audio.wav")

# 1. Transcribe
model = whisperx.load_model("tiny", device)
result = model.transcribe(audio, progress_callback=lambda p: print(f"Transcribe: {p:.1f}%"))

# 2. Align
align_model, metadata = whisperx.load_align_model(result["language"], device)
result = whisperx.align(result["segments"], align_model, metadata, audio, device, progress_callback=lambda p: print(f"Align: {p:.1f}%"))

# 3. Diarize
diarize_pipeline = DiarizationPipeline(token=hf_token, device=device)
diarize_df = diarize_pipeline(audio, progress_callback=lambda p: print(f"Diarize: {p:.1f}%"))
result = whisperx.assign_word_speakers(diarize_df, result)
```

### DIFF
--- a/whisperx/alignment.py
+++ b/whisperx/alignment.py
@@ -19,6 +19,7 @@ from whisperx.schema import (
     SingleAlignedSegment,
     SingleWordSegment,
     SegmentData,
+    ProgressCallback,
 )
 import nltk
 from nltk.data import load as nltk_load
@@ -122,6 +123,7 @@ def align(
     return_char_alignments: bool = False,
     print_progress: bool = False,
     combined_progress: bool = False,
+    progress_callback: ProgressCallback = None,
 ) -> AlignedTranscriptionResult:
     """
     Align phoneme recognition predictions to known transcription.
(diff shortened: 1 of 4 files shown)


julianpoemp:
for anyone wondering why the usage example outputs the progress to the console after the related function has been finished: you need to add the argument `flush=True`to the `print()` function. After that the callback writes to the console asap :)

PR #1391, opened 2026-03-27
Add Indonesian model to alignment.py

aziib:
support for indonesian language with safetensor model variant.


Barabazs (changes requested):
Thanks for adding Indonesian language support! However, could you point to the original model `cahya/wav2vec2-large-xlsr-indonesian` instead? The transformers library handles both pytorch and safetensors formats transparently, so there's no need for a separate safetensors re-upload.

aziib:
sure

Issue #1287, opened 2025-11-12
How to speedup alignment of whisperx?


type-shangshu:
I transcribed a 35min audio with whisperx, it takes about 41 seconds in `whisperx.align`; about 26 seconds in diarization. The total inference time is around 82 seconds, about 50% time spent on `whisperx.align`.
I noticed that there is TODO in alignment.py, (line 239, `# TODO: Probably can get some speedup gain with batched inference here`) Is there any plane to speedup alignment.py?

Issue #1222, opened 2025-09-03
GPUs utilization for Alignment and diarization


metheofanis:
I run whisperx on Lightning AI. I get some high performance GPUs.
During the transcription I get high GPU utilization and fast results. I use high --batch_size. 
The issue is with alignment and diarization. I get low GPU utilization, low VRAM usage and slower results.
Do I miss something? Is there any setting to help?
Thanks.

yysindi:
did you figure this this out? Diarization seems to be fine for me, but when i enable alignment my GPU util becomes very spiky

metheofanis:
I'm afraid no

yysindi:
Looks like there's a lot of parts of forced alignment that are CPU bound. also model is relatively small compared to transcription. if it helps what I did is i wrote my own custom implementation where I just split up the part that used GPU (ie wav2vec forward pass) and the parts that use CPU and ran them in two different stages of the pipeline to avoid the the latter part bottlenecking the GPU

Something like this:
First stage
- build waveform_segment
- run model(waveform_segment.to(cuda))
- store: emissions (ideally FP16 on disk to save space); stride / frame-to-time info (depends on model / feature extractor); the segment’s start, end, and cleaned text

Second stage (Later, on CPU):
- load emissions (as CPU tensor/np array)
- run get_trellis(...), backtrack(_beam)(...), merges, sentence split, dataframe smoothing

There's also some batching you can probably do to make it run faster.

metheofanis:
That's a great inspiration. Thanks a lot.

Issue #639, opened 2023-12-21
Alignement seems slow compared to transcription


brvier:
While transcribe take often less than 30s for transcription, alignment on the fr model take around 120s and sometime more than 200s for a 300s audio file.
Look like the GPU compute nothing, but the model is well loaded on GPU if i believe memory used on nvtop.

Is that alignment time is usual ?

brvier:
It s clearly visible in an nvtop graph each cpu to 0% are happening while align.
![2023-12-21_19-12-1703181712](https://github.com/m-bain/whisperX/assets/360591/afe5648b-b09e-4b20-a6ad-38f45d351e86)

zmy1116:
a significant portion of alignment is essentially performing dynamic programming,   you can try to optimize `get_trellis` and `backtrack` a little bit more...  otherwise torchaudio has a cuda implementation of the aligner,  you can try to integrate that [torchaudio.functional.forced_align()](https://pytorch.org/audio/2.1.2/generated/torchaudio.functional.forced_align.html#torchaudio.functional.forced_align)

Barabazs:
Closing this issue due to inactivity and its age.

If this issue persists, please feel free to reopen it.

PR #1396, opened 2026-03-31
feat: Rust extension for alignment DP

### LINKED ISSUE #219: improve alignment stage efficiency
1. alignment inference should be batched
2. [optionally] potentially use faster DTW algorithm / speed up pandas operations

ekzhang:
Hello! I noticed when using WhisperX in a service context that it would often spend over half of its time in the alignment phase, and also this phase would be mostly overhead from Python code in between CPU operations like `get_trellis` and `backtrack`. So it would take the GIL and degrade service latencies. (see #1222 #639 #1287, also this resolves #219)

We can speed this up a lot by just moving the CTC forced alignment into native code and dropping the GIL, as all of it is pure DP functions. I have a diff here that does exactly that — although, it adds an extension module with Rust.

Understand if this might not be the direction you want to take the project, just wanted to give an option!

cc @Barabazs @1carlito @m-bain 

---

_AI codegen attribution: This was all ported by AI with guidance by me_


Barabazs:
Hi @ekzhang first of all, thanks for contributing!

I'm not to keen on this approach tbh. A couple of thoughts:
* increased complexity by introducing another language and toolchain
* increased complexity for making releases (need platform-specific wheels)

I'm wondering if dropping pandas DataFrames in favor of plain dicts/numpy and possibly using forced_align from torchaudio directly would provide a similar improvement.

ekzhang:
Thanks for getting back, that's a fair criticism on the native code angle @Barabazs — I hear you that it would definitely require platform-specific wheels, and it makes development harder for people reading two languages.

(I did use the numpy bindings instead of torch so we don't depend on libtorch at least)

Dropping pandas DataFrames would make things a bit faster, but I did some profiling and found that most of the time was actually spent in the alignment and backward steps. [torchaudio's forced_align is deprecated and being removed from the next 2.9 release](https://docs.pytorch.org/audio/main/generated/torchaudio.functional.forced_align.html) unfortunately.

Not sure if there's an easy way to add native extensions without the extra build. I think I could do it in JAX with a scan targeting XLA, or maybe Numba? Cython would have the same platform-specific wheel issue. Will try and see.

PR #1400, opened 2026-04-03
Add Indonesian model to alignment.py

aziib:


Barabazs (approved):
Thank you for your contribution

PR #1407, opened 2026-04-16
Add token alias for Whisper model loading

### LINKED ISSUE #1406: Released wheel still uses use_auth_token in whisperx/asr.py with pyannote 4.x
## Summary

There appears to be one remaining `use_auth_token` call path in released WhisperX wheels, even though other pyannote-related code has already migrated to `token`.

In current released `whisperx==3.8.5`, `whisperx/vads/pyannote.py` uses `token=...`, but `whisperx/asr.py` still contains a `use_auth_token` parameter / call path in the VAD setup flow.

## Why this matters

`pyannote.audio` / Hugging Face usage has moved to `token`, and this remaining `use_auth_token` path appears to keep WhisperX only partially migrated.

That means downstream users still need to patch the installed wheel to avoid compatibility problems with pyannote 4.x.

## Evidence

From the released `whisperx==3.8.5` wheel:

- `whisperx/vads/pyannote.py` uses `token`
- `whisperx/asr.py` still contains:
  - `use_auth_token: Optional[Union[str, bool]] = None`
  - a downstream call passing `use_auth_token=use_auth_token`

In local validation against current released packages, this is the remaining call site that still requires a post-install patch.

## Expected behavior

WhisperX should consistently use `token` instead of `use_auth_token` across all pyannote-related code paths, including `whisperx/asr.py`.

## Suggested fix

Replace the remaining `use_auth_token` usage in `whisperx/asr.py` with `token`, so the migration is complete and released wheels no longer require downstream monkey-patching.

## Context

This looks related to the earlier pyannote 4 compatibility work, but I believe this specific remaining `asr.py` call path is still present in the released wheel and may not be fully resolved yet.

bobsummerwill:
## Summary

Adds a `token` alias to `whisperx.asr.load_model()` while preserving the existing `use_auth_token` argument for backward compatibility.

## Why

WhisperX has already migrated its pyannote-facing code to `token`, but the ASR model-loading surface still exposed `use_auth_token`.

The remaining internal use still needs to be forwarded to `faster-whisper`, whose current `WhisperModel` API expects `use_auth_token`.

This change updates the WhisperX API surface to accept `token` consistently without breaking the downstream faster-whisper call.

## Changes

- add `token` as the preferred alias in `whisperx.asr.load_model()`
- keep `use_auth_token` as a legacy alias for backward compatibility
- raise a `ValueError` if both are passed with conflicting values
- update `whisperx/transcribe.py` to pass `token=hf_token`

## Validation

- `python3 -m py_compile whisperx/asr.py whisperx/transcribe.py`

Fixes #1406

### DIFF
--- a/whisperx/asr.py
+++ b/whisperx/asr.py
@@ -327,6 +327,7 @@ def load_model(
     download_root: Optional[str] = None,
     local_files_only=False,
     threads=4,
+    token: Optional[Union[str, bool]] = None,
     use_auth_token: Optional[Union[str, bool]] = None,
 ) -> FasterWhisperPipeline:
     """Load a Whisper model for inference.
@@ -343,9 +344,16 @@ def load_model(
         download_root - The root directory to download the model to.
         local_files_only - If `True`, avoid downloading the file and return the path to the local cached file if it exists.
         threads - The number of cpu threads to use per worker, e.g. will be multiplied by num workers.
+        token - Hugging Face authentication token. Preferred alias.
+        use_auth_token - Legacy Hugging Face authentication token alias kept for backward compatibility.
     Returns:
         A Whisper pipeline.
     """
+
+    if token is not None and use_auth_token is not None and token != use_auth_token:
+        raise ValueError("Received conflicting values for 'token' and 'use_auth_token'")
+
+    token = token if token is not None else use_auth_token
 
     if compute_type == "default":
         compute_type = "float16" if device == "cuda" else "float32"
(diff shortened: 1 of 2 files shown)


Barabazs:
Closing, as noted in #1406, `use_auth_token` here is passed to faster-whisper's `WhisperModel`, which still expects that parameter name. Adding a `token` alias introduces indirection without functional benefit; we can revisit if/when faster-whisper migrates its API.

PR #1414, opened 2026-05-15
Fix Windows CUDA detection: include AMD64 in uv source platform markers

### LINKED ISSUE #1357: Incorrect torch and torchaudio source markers prevent CUDA installation on Windows
The current `tool.uv.sources` configuration uses `platform_machine == 'x86_64'` to identify GPU-capable systems. However, on Windows, `platform.machine()` returns `AMD64`.

```cmd
Python 3.12.12 (main, Dec  9 2025, 19:02:55) [MSC v.1944 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
Ctrl click to launch VS Code Native REPL
>>> import platform
>>> platform.system()
'Windows'
>>> platform.machine()
'AMD64'
```

**Problem:**
Currently, Windows users are forced into the pytorch-cpu index because AMD64 != x86_64.

**How to reproduce:**
- System: Windows 11
- Python: 3.12.12
-  Use the command: `uv sync` in the cloned project

**Proposed fix:**
Update markers to include `platform_machine == 'AMD64'` for Windows compatibility.

deekshaNVIDIA:
## Summary

Fixes #1357 — On Windows, `platform.machine()` returns `'AMD64'` instead of `'x86_64'`, causing the `[tool.uv.sources]` markers to always route `torch` and `torchaudio` to the `pytorch-cpu` index, even on CUDA-capable Windows systems.

This PR adds `platform_machine == 'AMD64'` alongside the existing `'x86_64'` check in the `torch` and `torchaudio` source markers, so Windows GPU users correctly resolve to the CUDA `pytorch` index.

## What changed

**`pyproject.toml`** — Updated the `[tool.uv.sources]` markers for `torch` and `torchaudio`:

- **GPU index marker**: `platform_machine == 'x86_64'` → `(platform_machine == 'x86_64' or platform_machine == 'AMD64')`
- **CPU fallback marker**: `platform_machine != 'x86_64'` → `platform_machine != 'x86_64' and platform_machine != 'AMD64'`

## Verification

On Windows 11 (Python 3.12):

```
>>> import platform
>>> platform.machine()
'AMD64'
```

Before this fix, `AMD64` did not match `x86_64`, so the CPU fallback rule matched instead of the GPU rule.

## Test plan

- [x] Confirmed `platform.machine()` returns `'AMD64'` on Windows
- [x] Verified the marker logic correctly routes to the GPU index for `AMD64`
- [x] macOS (`darwin`) routing is unchanged (first rule matches regardless of machine)
- [x] Linux `x86_64` routing is unchanged
- [x] Non-x86/non-AMD64 platforms (e.g., `aarch64` Linux) still fall through to CPU index

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -55,13 +55,13 @@ exclude-newer = "1 week"
 [tool.uv.sources]
 torch = [
   { index = "pytorch-cpu", marker = "sys_platform == 'darwin'" },
-  { index = "pytorch-cpu", marker = "platform_machine != 'x86_64' and sys_platform != 'darwin'" },
-  { index = "pytorch", marker = "platform_machine == 'x86_64' and sys_platform != 'darwin'" },
+  { index = "pytorch-cpu", marker = "platform_machine != 'x86_64' and platform_machine != 'AMD64' and sys_platform != 'darwin'" },
+  { index = "pytorch", marker = "(platform_machine == 'x86_64' or platform_machine == 'AMD64') and sys_platform != 'darwin'" },
 ]
 torchaudio = [
   { index = "pytorch-cpu", marker = "sys_platform == 'darwin'" },
-  { index = "pytorch-cpu", marker = "platform_machine != 'x86_64' and sys_platform != 'darwin'" },
-  { index = "pytorch", marker = "platform_machine == 'x86_64' and sys_platform != 'darwin'" },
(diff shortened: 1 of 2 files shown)


ultrasound1372:
I was about to open a PR to do exactly the same thing which is confirmed working. +1.

Barabazs (approved):

Barabazs:

PR #1419, opened 2026-05-23
fix(alignment): handle 'ignore' interpolation method gracefully (#1368)

### LINKED ISSUE #1368: Interpolation method "ignore" raises an error
~When using the latest version of Pandas (3.0.0)~, the interpolation method "ignore" raises a value error. 
**EDIT**: it is independent of the Pandas version (tested it with `2.1`, `2.2`, `2.3`, and `3.0`) 

Image

For what I see, there is nowhere in the WhisperX code where the "ignore" method is treated as an exception**, but [it's not in the Pandas method](https://pandas.pydata.org/pandas-docs/version/2.3/reference/api/pandas.DataFrame.interpolate.html). Instead, it seems to be passed directly to Pandas: https://github.com/m-bain/whisperX/blob/064f7379981d744ae14ceedf92c3d1e956db18a9/whisperx/utils.py#L470-L474 

** Actually, checking the entire codebase, it only appears once, in [line 34](https://github.com/m-bain/whisperX/blob/636f2988f82bcb3ad2450c4f85865a8d435a0b7c/whisperx/__main__.py#L34) of `__main__.py`

```python
    parser.add_argument("--interpolate_method", default="nearest", choices=["nearest", "linear", "ignore"], help="For word .srt, method to assign timestamps to non-aligned words, or merge them into neighbouring.")
```

awf-cpu:
Closes #1368

## Problem
When using the '--interpolate_method ignore' option (which is a valid argparse option), pandas raises a ValueError ('Can not interpolate with method=ignore') because pandas' interpolate() function doesn't support the 'ignore' method.

## Fix
In 'interpolate_nans()', bypass the pandas interpolate() call when the method is set to 'ignore', falling back to forward and backward fill (ffill/bfill) directly.

## Tests
Added 'test_interpolate_method_ignore' to 'tests/test_word_timestamp_interpolation.py' to verify that 'ignore' works without throwing ValueError and correctly yields timestamps.

### DIFF
--- a/tests/test_word_timestamp_interpolation.py
+++ b/tests/test_word_timestamp_interpolation.py
@@ -61,7 +61,7 @@ class TestAlignWithWildcards:
     }
     METADATA = {"language": "en", "dictionary": DICTIONARY, "type": "torchaudio"}
 
-    def _run_align(self, text, duration=5.0, num_frames=100):
+    def _run_align(self, text, duration=5.0, num_frames=100, interpolate_method="nearest"):
         """Run align() with a mock model on a single segment."""
         torch.manual_seed(0)
         emission = _make_emission(num_frames, self.DICTIONARY, list(text), blank_id=0)
@@ -78,6 +78,7 @@ class TestAlignWithWildcards:
             align_model_metadata=self.METADATA,
             audio=audio,
             device="cpu",
+            interpolate_method=interpolate_method,
         )
         return result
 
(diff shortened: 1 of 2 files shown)


Barabazs:
You're still falling back to ffill().bfill() when method is "ignore", which still fills in timestamps for non-aligned words, just with a different strategy.

PR #1436, opened 2026-06-14
[codex] document uv workflow

cinnabarhorse:
## Summary
- add AGENTS.md guidance requiring uv for Python dependency, environment, and test commands
- pin .python-version to 3.12 so uv does not select the local Intel Python 3.10 on Apple Silicon

## Root cause
The project already uses uv and PyTorch-specific uv indexes, but the tracked .python-version was 3.10. On this machine uv selected /usr/local Python 3.10 x86_64, which cannot install the macOS arm64-only torch 2.8.0 CPU wheel.

## Validation
- uv sync --all-extras --dev
- uv run pytest
- uv run pytest tests/ -v
- uv run python -c "import whisperx; print('Successfully imported whisperx')"


cinnabarhorse:
Closing because this branch should only be pushed to the fork, not opened against upstream.

PR #1441, opened 2026-06-21
fix: skip initial silence to improve language detection accuracy

yccheok:
Previously, `detect_language` always analyzed the first 30 seconds of the raw audio file. If a file started with a long period of silence, the language detector would evaluate background noise, resulting in inaccurate language guesses.

This change leverages the already-computed `vad_segments` to find the timestamp of the first actual spoken word. By slicing the audio array at `first_speech_start_idx`, the language detector now correctly analyzes the first 30 seconds of actual speech. A fallback remains in place for files where no speech is detected.

### DIFF
--- a/whisperx/asr.py
+++ b/whisperx/asr.py
@@ -216,14 +216,26 @@ class FasterWhisperPipeline(Pipeline):
             offset=self._vad_params["vad_offset"],
         )
         if self.tokenizer is None:
-            # If VAD found speech segments, start language detection from the first spoken word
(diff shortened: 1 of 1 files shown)


yccheok:
For testing purpose, I include a voice file for testing.

The language is in Chinese.

Detecting language starting at 37.85s

[skip-initial-silence.m4a.zip](https://github.com/user-attachments/files/29175250/skip-initial-silence.m4a.zip)

PR #1469, opened 2026-08-27
Make torchaudio optional with Hugging Face fallback for align models

li-lizhe:
## Problem

`torchaudio` is a hard dependency of whisperX, but PyTorch publishes no Linux
aarch64 wheels for torchaudio. On Ascend NPU / Kunpeng ARM servers (whose torch
is version-decoupled from upstream PyTorch), `pip install whisperx` fails
outright, and even a source install can't import `whisperx.alignment` because
`import torchaudio` sits at module top-level.

## Change

Make torchaudio optional while keeping behavior unchanged when it *is* available:

- Guard `import torchaudio` with `try/except ImportError`, so alignment imports without it.
- Add `TORCHAUDIO_PIPELINE_TO_HF`, mapping the five `torchaudio.pipelines` align
  models to their equivalent Hugging Face checkpoints (e.g. `WAV2VEC2_ASR_BASE_960H`
  -> `facebook/wav2vec2-base-960h`). When torchaudio is absent, the default
  en/fr/de/es/it models fall back to these HF checkpoints.
- Move `torchaudio` from `dependencies` to an optional `align-torch` extra in
  `pyproject.toml` (only needed for the torchaudio.pipelines path).

## Verification

On an Ascend 910B / aarch64 container with **no torchaudio installed**
(torch 2.14.0a0 + torch_npu):

- `import whisperx.alignment` succeeds; module-level `torchaudio` is `None`.
- `TORCHAUDIO_PIPELINE_TO_HF` maps all 5 default torch models to `facebook/...` checkpoints.
- All 36 HF-default languages are unaffected.

Added `tests/test_align_model_fallback.py` to guard the mapping against regressions.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,7 +18,6 @@ dependencies = [
     "pyannote-audio>=4.0.0",
     "huggingface-hub>=0.28.1",
     "torch~=2.8.0",
-    "torchaudio~=2.8.0",
     "torchvision~=0.23.0",
     "torchcodec>=0.6.0,<0.8.0; (sys_platform == 'linux' and platform_machine == 'x86_64') or sys_platform == 'darwin' or sys_platform == 'win32'",
     "transformers>=4.48.0",
@@ -28,6 +27,9 @@ dependencies = [
 
 [project.optional-dependencies]
 dev = ["pytest"]
+# torchaudio 仅用于 torchaudio.pipelines 加载对齐模型（可选路径）。
+# 无需 torchaudio 的平台（如昇腾/ARM aarch64，无 torchaudio wheel）默认不安装它。
+align-torch = ["torchaudio~=2.8.0"]
 
 [project.scripts]
 whisperx = "whisperx.__main__:cli"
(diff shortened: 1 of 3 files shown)


li-lizhe:
Thanks for the review @copilot-pull-request-reviewer — all points addressed in 51ca65b:

1. **Import guard now catches any `Exception`** (not just `ImportError`), so broken torchaudio installs (missing native libs, ABI mismatch) degrade to the HuggingFace path instead of crashing `import whisperx.alignment`.
2. **Comments translated to English** in `whisperx/alignment.py` and `pyproject.toml` for consistency with the rest of the codebase.
3. **Added a direct behavioral test** (`test_load_align_model_falls_back_to_hf_when_torchaudio_missing`): with `torchaudio` patched to `None`, `load_align_model("en")` remaps `WAV2VEC2_ASR_BASE_960H` → `facebook/wav2vec2-base-960h` and loads it via the `Wav2Vec2Processor`/`Wav2Vec2ForCTC` path, returning `type="huggingface"` metadata.

Verified on Ascend 910B (Linux aarch64, torch 2.14.0a0 + torch_npu, no torchaudio installed): all 3 tests in `tests/test_align_model_fallback.py` pass.

PR #1470, opened 2026-08-27
feat: add batch context, batch interleaving, and redo first batch

1carlito:
## Summary
Adds two opt-in flags that improve transcription coherence when using batched inference — zero impact on default behaviour.

## Changes
- **`whisperx/asr.py`**
  - `generate_segment_batched`: per-stream prompt injection using rolling token context
  - `FasterWhisperPipeline._forward`: accumulates output tokens per stream after each batch
  - `FasterWhisperPipeline.transcribe`: batch interleaving of VAD segments across N streams; wrap-around redo of first batch; final sort to restore chronological order
  - `load_model`: exposes `use_batch_context` param

- **`whisperx/transcribe.py`**: wires `--batch_context` and `--redo` through `transcribe_task`

- **`whisperx/__main__.py`**: adds `--batch_context` and `--redo` CLI flags

## Usage
```bash
# Batch context only (interleaving + rolling context)
whisperx audio.mp3 --model large-v2 --batch_size 8 --batch_context

# Batch context + redo (wrap-around context on first batch)
whisperx audio.mp3 --model large-v2 --batch_size 8 --batch_context --redo


All new flags default to False. A user running whisperX without these flags gets identical behaviour to upstream — no code paths are entered, no performance impact.

### DIFF
--- a/whisperx/__main__.py
+++ b/whisperx/__main__.py
@@ -77,6 +77,7 @@ def cli():
     parser.add_argument("--hf_token", type=str, default=None, help="Hugging Face Access Token to access PyAnnote gated models")
 
     parser.add_argument("--print_progress", type=str2bool, default = False, help = "if True, progress will be printed in transcribe() and align() methods.")
+    parser.add_argument("--redo", action="store_true", help="Re-run the first batch of streams with wrap-around context")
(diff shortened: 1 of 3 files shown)


m-bain:
Remove redo option just make it always do redo as per the paper 

Also can you add a new warning if batch size is similar or equal to total number chunks; advising user to reduce batch size

PR #1471, opened 2026-08-28
feat: add batch interleaving and cold start resolution 

1carlito:
Summary
Adds a new opt-in flag that improve transcription coherence when using batched inference — zero impact on default behaviour.

Changes
whisperx/asr.py

generate_segment_batched: per-stream prompt injection using rolling token context
FasterWhisperPipeline._forward: accumulates output tokens per stream after each batch
FasterWhisperPipeline.transcribe: batch interleaving of VAD segments across N streams; wrap-around redo of first batch; final sort to restore chronological order

whisperx/transcribe.py: wires --redo through transcribe_task

whisperx/__main__.py: adds --redo CLI flag

Usage

# redo (interleaving, rolling context and wrap-around context on first batch)
e.g whisperx audio.mp3 --model large-v2 --batch_size 8  --redo

New flag defaults to False. A user running whisperX without this flag gets identical behaviour to upstream — no code paths are entered, no performance impact.

Warning added:
if batch size is similar or equal to the total number chunks, user is advised to reduce batch size.

### DIFF
--- a/whisperx/__main__.py
+++ b/whisperx/__main__.py
@@ -77,6 +77,7 @@ def cli():
     parser.add_argument("--hf_token", type=str, default=None, help="Hugging Face Access Token to access PyAnnote gated models")
 
     parser.add_argument("--print_progress", type=str2bool, default = False, help = "if True, progress will be printed in transcribe() and align() methods.")
+    parser.add_argument("--redo", action="store_true", help="Re-run the first batch of streams with wrap-around context")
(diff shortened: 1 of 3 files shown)


m-bain:
why is the flag just `--redo` now, i think lets just use `--condition_on_previous_text` or `--interleaved_context` flag and redo should happen automatically (not a flag) when this is used. Redo is part of the method lets not make it configurable

